PR #1, opened 2022-09-10
internal: remove a couple minor stale/unused pieces

euank:
Notably, remove the very old TODO files.

Also, do some minor shuffling around accuracy of the lib/tunnel/proto stuff, including dropping the slightly spooky sounding docs for the unused fingerprint field.


internal/muxado/TODO
```diff
@@ -1,32 +0,0 @@
...
-More session tests
-More stream tests
-frame/framer tests - return proper error types, hand unknown type frames
-
-### Low priority:
-extension: Move high throughput connections to their own connections
-window_manager needs to support a deadline
-inboundbuffer deadline support
-add Reset() stream API?
-eliminate unlikely race on s.remoteDebug between handleFrame() and die()
```
euank, line 31: By the by, this TODO is fixed, isn't it :stuck_out_tongue:

jrobsonchase (approved):
:+1: 

Also adding a tag to the PR so we can keep track of things that need to be side-ported to libngrok until we can unify the codebases somehow

euank:
Updated to be almost the same PR, but to retain the fingerprint field.

jrobsonchase (approved):

PR #2, opened 2022-09-12
Make the 'log' submodule dep optional

euank:
Currently, trying to actually depend on libngrok requires getting two modules, the main one, and the 'log' one.

Go modules make this have confusing UX!

Right now, this UX is broken! We would have needed to set a more specific version in the `go.mod` for the log dependency (i.e. v0.0.1 after tagging a release).

The current hoops we have to jump through to use this library are:

```
require github.com/ngrok/libngrok-go v0.0.0-20220909154412-fe6c40f2e674

replace github.com/ngrok/libngrok-go/log v0.0.0 => github.com/ngrok/libngrok-go/log v0.0.0-20220909154412-fe6c40f2e674
```

and that replace isn't given to you by the go tools, you have to find it yourself.

This updates how logging is structured to copy the log interface into both locations, which due to go's structural typing should work for our needs, but also importantly makes pulling in this code as a dependency easier and less confusing.

To quote the rob pike proverb: "A little copying is better than a little dependency."

Notably, this makes a naive `go get` just work, without having to find any of the gunk above.


flake.lock
```diff
@@ -2,11 +2,11 @@
...
     "flake-utils": {
       "locked": {
-        "lastModified": 1656928814,
-        "narHash": "sha256-RIFfgBuKz6Hp89yRr7+NR5tzIAbn52h8vT6vXkYjZoM=",
+        "lastModified": 1659877975,
+        "narHash": "sha256-zllb8aq3YO3h8B/U0/J1WBgAL8EX5yWf5pMj3G0NAmc=",
         "owner": "numtide",
         "repo": "flake-utils",
-        "rev": "7e2a3b3dfd9af950a856d66b0a7d01e3c18aa249",
+        "rev": "c0e246b9b83f637f4681389ecabcb2681b4f3af0",
```
euank, line 9: `nix flake update` which snuck in while I was trying to get my doc link to render.

jrobsonchase (approved):
Whoops, that's what I get for trying to be clever. Thanks for fixing this!

PR #4, opened 2022-09-15
assets: move to a subdirectory and update CA cert

jrobsonchase:
Small amount of cleanup - create a new directory in which to store static assets, put our root CA cert there, and update it to the latest version.


euank (approved):
:+1:

PR #5, opened 2022-09-15
scripts: relocate Makefile to scripts directory

jrobsonchase:
Still to be seen if this will remain useful for CI/convenience purposes, or if we'll end up removing it entirely.

For now, clean up the top level a bit and put it in a subdirectory.


euank (approved):
lgtm; I also have no issue with dropping it until we do find a CI need

jrobsonchase:
I still use it myself as a shortcut for commands I don't want to remember, or at the very least to jog my memory. Space is limited up there, gotta free it up somehow :stuck_out_tongue:

PR #6, opened 2022-09-15
session: clean up the public API

jrobsonchase:
Resolves a few product asks:

* Trims the `Session` interface down to just `Close` and `StartTunnel`
* Stops leaking `proto` types into the public API
* Hides the `ReconnectCookie`, making it an internal implementation detail

Most of the removed functionality is actually just hidden - it can still be accessed via interface assertions since the non-exported implementer of the `Session` interface still has the methods. I figure this should be sufficient if we ever find that we actually do need them internally for something.

### DIFF
--- a/session.go
+++ b/session.go
@@ -28,13 +28,6 @@ type Session interface {
 	Close() error
 
 	StartTunnel(ctx context.Context, cfg TunnelConfig) (Tunnel, error)
-
-	SrvInfo() (SrvInfo, error)
-	AuthResp() AuthResp
-
-	Heartbeat() (time.Duration, error)
-
-	Latency() <-chan time.Duration
 }
 
 const (
--- a/session.go
+++ b/session.go
@@ -324,8 +324,15 @@ func Connect(ctx context.Context, cfg *ConnectConfig) (Session, error) {
 		}
 
 		session.setInner(&sessionInner{
-			Session:  sess,
-			AuthResp: resp,
+			Session:         sess,
+			Region:          resp.Extra.Region,
+			ProtoVersion:    resp.Version,
+			ServerVersion:   resp.Extra.Version,
+			ClientID:        resp.Extra.Region,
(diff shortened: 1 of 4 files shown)


session.go
```diff
@@ -457,22 +456,37 @@ func (s *sessionImpl) StartTunnel(ctx context.Context, cfg TunnelConfig) (Tunnel
...
+func (s *sessionImpl) PlanName() string {
+	return s.inner().PlanName
+}
+func (s *sessionImpl) Banner() string {
+	return s.inner().Banner
+}
+func (s *sessionImpl) SessionDuration() int64 {
+	return s.inner().SessionDuration
+}
+func (s *sessionImpl) Region() string {
```
euank, line 484: region feels potentially interesting for anyone who uses the geo-balanced dns name.

Is that info available with a better ux somewhere?

euank (approved):
lgtm, one idle comment

session.go
jrobsonchase: Yeah, I agree. I'm just not sure where it would go.

I can think of a few approaches, none of which are great:
* Add some of these getters to the `Session` interface
* Add an `Info` method to the `Session` interface that returns a collection of these properties
* Return a concrete type rather than the interface, and expose them there

The first two dirty up the interface and are counter to the "Just `Close` and `StartTunnel`" goal, and the later makes it harder for us to "hide" things from the public API, but still keep them kinda accessible.

euank: `Session.Metadata()` doesn't seem _that bad_.

But also, we could add it as an `OnConnect` interceptor, i.e. have that pass in `Session, SessionMetadata` as arguments. It's fixed for a session, so it can just be yanked out there reasonably.

I'm happy merging this as-is and iterating further if the need arises.

jrobsonchase: An interceptor argument seems interesting!

I'll make a note of these ideas for discussion next time we get everyone in a room for a review.

PR #7, opened 2022-09-15
tls: configuration cleanup

jrobsonchase:
This:

* Lifts the cert and key fields out of a sub-struct and into the main `TLSConfig`
* Renames `WithEdgeTermination` to `WithTermination`, since it's *always* at the edge.


euank (approved):
:+1:

PR #8, opened 2022-09-16
tunnel: add Close to top-level interface

jrobsonchase:
Adds `Close` to the top-level `Tunnel` interface.

Also changes it to call `CloseWithContext` with a default 5s timeout internally (even though we don't actually use that yet).

### DIFF
--- a/tunnel.go
+++ b/tunnel.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"net"
 	"net/http"
+	"time"
 
 	tunnel_client "github.com/ngrok/libngrok-go/internal/tunnel/client"
 )
 
 type Tunnel interface {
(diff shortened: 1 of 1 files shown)


tunnel.go
```diff
@@ -4,12 +4,14 @@ import (
...
 	"net"
 	"net/http"
+	"time"
 
 	tunnel_client "github.com/ngrok/libngrok-go/internal/tunnel/client"
 )
 
 type Tunnel interface {
 	CloseWithContext(context.Context) error
+	Close() error
```
euank, line 14: This pair of functions could use a doc-comment to explain when you might want one or the other.

euank (approved):
lgtm, doc comments would be nice

PR #9, opened 2022-09-16
examples: add dynamic

jrobsonchase:
This is what I showed off internally on Monday.

It uses a pretty naiive HTTP api to start/stop tunnels that do nothing but do oauth and echo the request back.

Builds a little bit on #8, so we should get that merged first.

### DIFF
--- a/example/dynamic.go
+++ b/example/dynamic.go
@@ -88,7 +88,7 @@ func manageTunnels(ctx context.Context, sess libngrok.Session) http.Handler {
 				WithOAuth(libngrok.OAuthProvider(provider).
 					AllowEmail(allowed)))
 			if err != nil {
-				httpResp(w, http.StatusInternalServerError, "error starting tunnel\n")
+				httpResp(w, http.StatusInternalServerError, "error starting tunnel: %v\n", err)
 				return
 			}
 
--- a/example/dynamic.go
+++ b/example/dynamic.go
@@ -1,0 +1,140 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"log"
+	"net/http"
+	"net/http/httputil"
+	"net/url"
+	"os"
+	"strings"
+	"sync"
+
+	"github.com/inconshreveable/log15"
+	"github.com/ngrok/libngrok-go"
+)
+
+func unwrap[T any](out T, err error) T {
+	if err != nil {
+		log15.Error("unwrapped error", "error", err)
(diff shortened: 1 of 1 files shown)


example/dynamic.go
```diff
@@ -0,0 +1,140 @@
...
+				OnRestart: func(ctx context.Context, sess libngrok.Session) error {
+					log15.Info("restarting due to remote request")
+					return nil
+				},
+			}).
+			WithAuthToken(os.Getenv("NGROK_TOKEN")),
+		))
+
+		tun := unwrap(sess.StartTunnel(ctx, libngrok.HTTPOptions().
+			WithDomain(domain).
```
alex-bezek, line 58: this is an empty string when it first is ran, is this just so if it restarts on accident you re-use the same random domain name you got the first time?

example/dynamic.go
```diff
@@ -0,0 +1,140 @@
...
+		if r.Method == http.MethodGet && r.URL.Path == "/create" {
+			provider := r.FormValue("provider")
+			allowed := r.FormValue("allow")
+			if provider == "" || allowed == "" {
+				httpResp(w, http.StatusBadRequest, "must supply 'provider' and 'allow' form values\n")
+				return
+			}
+
+			tun, err := sess.StartTunnel(ctx, libngrok.HTTPOptions().
+				WithForwardsTo("dump requests").
```
alex-bezek, line 87: what is this "dump requests" string? The underlying function says it should be an addr? https://github.com/ngrok/ngrok-go/blob/0703564d4eb34c167faa4b3e2e2af9da4de358d2/labeled.go#L23

alex-bezek (approved):

example/dynamic.go
euank: This isn't actually used for anything other than showing up in the ngrok dashboard/api as "this tunnel is forwarding to X"

In typical tunnels, if you run `ngrok http http://my-local-service:8080`, that final string is what gets plumbed up. The agent normalizes the address some (so iirc `ngrok http 80` is `ForwardsTo: http://localhost:80` or similar).

euank (approved):

example/dynamic.go
jrobsonchase: Yeah, this only makes sense with the reconnect cookie, which we're removing from the public API.

So the domain arg here will go away as well.

PR #11, opened 2022-09-19
examples: Restructure, standardize, and add more

jrobsonchase:
WIP, more to come


jrobsonchase:
Gonna go ahead and rebase/mark this ready to go. We can add more examples as we think of specific cases.

inconshreveable:
1. Let's avoid any dependencies except the standard library. Replace log15 with Printf statements, replace spew with Printf statements as well, probably.
2. Avoid any 'common' library for examples and just copy/paste that code or functions into the relevant examples. We want developers to be able to copy paste our examples anywhere and import our library and go.
3. Let's get some much simpler examples to start with. Many examples which each highlight one specific feature are often preferable to one kitchen sink example.

here's the simplest example i'm using currently:

```
package main

import (
	"context"
	"log"
	"net/http"
	"os"

	ngrok "github.com/ngrok/libngrok-go"
)

func main() {
	if err := run(context.Background()); err != nil {
		log.Fatal(err)
	}
}

func run(ctx context.Context) error {
	_, tun, err := ngrok.ConnectAndStartTunnel(ctx,
		ngrok.ConnectOptions().
			WithAuthtoken(os.Getenv("NGROK_AUTHTOKEN")),
		ngrok.HTTPOptions().
			WithDomain("golang-demo.ngrok.io"),
	)
	if err != nil {
		return err
	}
	return http.Serve(tun.AsListener(), http.HandlerFunc(handler))
}

func handler(w http.ResponseWriter, r *http.Request) {
	w.Write([]byte("Hello from ngrok-go!\n"))
}
```

jrobsonchase:
@inconshreveable 

1. Done! Just a `golang.org/x/sync` dependency left, but I figure those are "std++" and fairly well accepted.
2. Aaaand it's gone
3. There are now four examples: `http` (your demo more or less), `tcp`, `logging`, and `ngrok-lite`

I rebased it on the "rebrand" branch for now to avoid more name thrashing until more decisions get made.

examples/ngrok-lite/main.go
```diff
@@ -0,0 +1,77 @@
+// Naiive ngrok agent implementation.
```
inconshreveable, line 1: naiive -> naive

examples/http/main.go
```diff
@@ -0,0 +1,36 @@
...
+		log.Fatal(err)
+	}
+}
+
+func run(ctx context.Context) error {
+	_, tun, err := ngrok.ConnectAndStartTunnel(ctx,
+		ngrok.ConnectOptions().
+			WithAuthtoken(os.Getenv("NGROK_AUTHTOKEN")),
+		ngrok.HTTPOptions().
+			WithDomain("golang-demo.ngrok.io"),
```
inconshreveable, line 26: let's remove this and add another example that uses domain selection with a dynamically chosen name. the hello world example should work on the free tier and it needs to be dynamic because any hardcoded name will get grabbed

inconshreveable: random domain also means the program should print it out

inconshreveable (approved):

inconshreveable:
we discussed further improvements offline, but approved for now and @russorat will take over the next phase of example work

PR #12, opened 2022-09-20
configs: factor out struct embedding, generics, and standardize receiver

jrobsonchase:
Simplifies config structures and makes it far clearer what methods are available where.


euank (approved):
:+1:; lgtm

PR #13, opened 2022-09-20
configs: always copy the config structures when calling one of the methods

jrobsonchase:
WIP-ish - still not sure if this is the right move. Builds on #12.

Modifying the receiver *and* returning it combines a couple of approaches to the builder/method chaining pattern, which is kinda weird.

This makes it so that all of the methods take a value receiver and perform an implicit copy that then gets returned. This is fine-ish, and makes it so that the zero values are actually useful, which may remove the need for the config constructor methods, but is something of a false promise when map or slice fields are modified. We could force a copy of those, but that ends up getting noisy with dubious benefits.

### DIFF
--- a/common.go
+++ b/common.go
@@ -7,7 +7,7 @@ import (
 )
 
 type CommonConfig struct {
-	CIDRRestrictions *CIDRRestriction
+	CIDRRestrictions CIDRRestriction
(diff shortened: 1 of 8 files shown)


jrobsonchase:
Going with functional options instead, this one is moot.

PR #14, opened 2022-09-21
tunnels: auto-define ForwardsTo

jrobsonchase:
Right now it's in the format `app://<hostname>/<executable path>?pid=<process id>`.

We can bikeshed here.

Builds on #12, which should be merged first.

### DIFF
--- a/common.go
+++ b/common.go
@@ -1,7 +1,10 @@
 package libngrok
 
 import (
+	"fmt"
 	"net"
+	"os"
+	"path/filepath"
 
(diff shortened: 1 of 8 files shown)


euank:
The diff for this one seems to still include #12; mind rebasing to hopefully make the diff smaller / easier to review?

jrobsonchase:
Weird - that usually auto-updates, doesn't it?

inconshreveable (approved):

PR #15, opened 2022-09-21
errors: remove contexts, generics, and type aliases

jrobsonchase:
Simplifies errors to a handful of concrete types.

More boilerplate, but nicer for consumers.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -6,74 +6,93 @@ import (
 	"reflect"
 )
 
-type ErrContext interface {
-	message() string
-}
-type Error[C ErrContext] struct {
-	Inner   error
-	Context C
+type ErrAuthFailed struct {
+	Remote bool
+	Inner  error
 }
 
-func (e Error[C]) Unwrap() error {
+func (e ErrAuthFailed) Error() string {
+	var msg string
+	if e.Remote {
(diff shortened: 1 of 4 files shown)


errors.go
```diff
@@ -6,74 +6,93 @@ import (
...
 
-func (e Error[C]) Error() string {
-	msg := e.Context.message()
-	if e.Inner != nil {
-		return fmt.Sprintf("%s: %v", msg, e.Inner.Error())
-	} else {
-		return msg
-	}
+func (e ErrAuthFailed) Is(target error) bool {
+	return reflect.TypeOf(e) == reflect.TypeOf(target)
```
euank, line 30: Since this is a concrete type, can't we do:

```
_, ok := target.(ErrAuthFailed)
return ok
```

instead?

Since the concrete type is known at compile-time, it seems like the checked type-assertion should be functionally the same, but faster.

Same comment for the other places using `reflect.TypeOf`

euank (commented):
Looks good overall, one comment about the `Is` impls

errors.go
jrobsonchase: Ope, right you are. Will fix

euank (approved):
:+1:

PR #19, opened 2022-09-23
callbacks: hide OnUpdate from the public API

jrobsonchase:
Also switches the callback handler to an interface in support of this. A struct implementing the public interface is provided to reduce boilerplate if all consumers need is a place to stick standalone functions.

### DIFF
--- a/session.go
+++ b/session.go
@@ -74,10 +74,42 @@ type RemoteCallbacks struct {
 	// It is the implementer's responsibility to ensure that the application
 	// recreates the session.
 	OnRestart func(ctx context.Context, sess Session) error
+}
+
+func (cb RemoteCallbacks) HandleStop(ctx context.Context, sess Session) error {
+	if cb.OnStop != nil {
+		return cb.OnStop(ctx, sess)
+	}
+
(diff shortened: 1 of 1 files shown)


session.go
```diff
@@ -74,10 +74,42 @@ type RemoteCallbacks struct {
...
+func (cb RemoteCallbacks) HandleRestart(ctx context.Context, sess Session) error {
+	if cb.OnRestart != nil {
+		return cb.OnRestart(ctx, sess)
+	}
+
+	return errors.New("restart callback not set")
+}
+
+type RemoteCallbackHandler interface {
+	// Called when a stop is requested via the dashboard or API.
```
euank, line 96: Nit, idiomatically should be documented as "HandleStop is called ...."

Same for the other godoc comments

euank (approved):
I know that there might be more changes incoming re the HandleX interface setup, but this still seems like a quite reasonable improvement if you want to merge this in the meanwhile!

jrobsonchase:
The configuration API is going to be changing in other significant ways - I don't think this is relevant anymore.

PR #20, opened 2022-09-23
Document ALL the things!

jrobsonchase:
Lots o' doc comments.

Also some misc. fixes/ergonomics improvements to a few configuration methods. Could split them out, but then there would be some back-and-forth on some of the comments since the behavior changes ever so slightly. Ditto on the auto-ForwardsTo change that I accidentally built this on top of :stuck_out_tongue:


common.go
```diff
@@ -1,66 +1,95 @@
...
 type CommonConfig struct {
+	// Restrictions placed on the origin of incoming connections to the edge.
 	CIDRRestrictions *CIDRRestriction
-	ProxyProto       ProxyProtoVersion
-	Metadata         string
-	ForwardsTo       string
+	// The version of PROXY protocol to use with this tunnel, zero if not
+	// using.
+	ProxyProto ProxyProtoVersion
+	// Tunnel-specific opaque metadata. Viewable via the API.
```
euank, line 18: This comment basically applies to all of the godoc comments in this PR:

The idiomatic styling for a godoc comment is to start with the thing it's documenting, iiuc, so for this one it would be:

"Metadata is tunnel-specific opaque metadata. Viewable via the API."

Quoting from https://go.dev/blog/godoc

> Notice this comment is a complete sentence that begins with the name of the element it describes

euank (commented):
It says "ALL the things", but there's one thing more we can pick!

We can do a `doc.go` containing `Package libngrok .....` with a package level doc comment.

Broadly, looks great! I'll give it another round after rewording for the go convention above.

common.go
jrobsonchase: Ugh, I've never liked this convention :roll_eyes: 

For all of the try-to-avoid-stuttering in the go ecosystem, it feels awfully similar. The guidelines make it sound like it's to prevent ambiguity when the final medium doesn't actually pair the comment with the thing it's documenting, but I've never seen one where that's actually a problem. /rant

Personal feelings aside, I guess we should play nice and get in line :stuck_out_tongue:

inconshreveable (approved):

common.go
jrobsonchase: Discussed with @russorat and @inconshreveable  - we're going to merge for now and make another pass at these later in the interest of reducing merge conflicts as other stuff is being worked on.

PR #21, opened 2022-09-26
ConnectAndStartTunnel no longer takes pointer receiver + add test

inconshreveable:


jrobsonchase (approved):
Seems reasonable to me :+1:

I've also got a draft that gives all of the config structs a similar treatment wrt all of the `With` methods: https://github.com/ngrok/ngrok-go/pull/13. Seems in some ways more right, and in other ways less right. Would appreciate some input there if you get a chance to look at it.

PR #24, opened 2022-10-06
session: remove support for custom 'not supported' errors

jrobsonchase:


session.go
```diff
@@ -272,20 +259,10 @@ func Connect(ctx context.Context, cfg *ConnectConfig) (Session, error) {
 	}
 
 	empty := ""
-	notImplemented := "not implemented"
+	notImplemented := "the agent does not support this command"
```
inconshreveable, line 262: "the agent has not defined a callback for this operation"

?

inconshreveable (approved):

PR #25, opened 2022-10-06
session: rework callback handlers

jrobsonchase:
They're now all individual fields rather than big collections. Better granularity and stuff.


jrobsonchase:
@inconshreveable Let's go ahead and get this merged.

The ConnectHandler still needs to become PostAuthHandler or something, and some of it will be superseded by the struct -> functional options transition, but this is still useful in the interim.

inconshreveable (approved):

PR #26, opened 2022-10-07
switch to functional options pattern for tunnel configuration

jrobsonchase:
Per multiple discussions, switch the tunnel configurations from structs-with-chainable-methods to opaque option-functions.

The connect options will be getting similar treatment before long.


errors.go
```diff
@@ -55,8 +55,6 @@ func (e ErrAcceptFailed) Is(target error) bool {
 
 // Errors arising from a failure to start a tunnel.
 type ErrStartTunnel struct {
-	// The configuration used to start the tunnel.
-	Config TunnelConfig
```
jrobsonchase, line 59: These are no longer inspectable

PR #35, opened 2022-10-07
embed net.Listener

inconshreveable:
every Tunnel is-a net.Listener. also reordered the methods in the interface so more relevant ones come first

### DIFF
--- a/tunnel.go
+++ b/tunnel.go
@@ -11,14 +11,9 @@ import (
 
 // An ngrok tunnel.
 type Tunnel interface {
-	// Closing a tunnel is an operation that involves sending a "close" message
-	// over the existing session. Since this is subject to network latency,
-	// packet loss, etc., it is most correct to provide a context. See also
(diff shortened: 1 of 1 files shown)


tunnel.go
```diff
@@ -38,16 +33,17 @@ type Tunnel interface {
...
 	AsHTTP() HTTPTunnel
-}
 
-// A tunnel that also implements [net.Listener].
-type ListenerTunnel interface {
-	Tunnel
-	net.Listener
+	// Convenience method that calls `CloseWithContext` with a default timeout
+	// of 5 seconds.
+	Close() error
```
jrobsonchase, line 41: Can we get rid of this since it's included in `net.Listener`, or is it here just for the doc comment?

jrobsonchase (approved):

tunnel.go
inconshreveable: yeah just here for the doc comment which seems important enough to leave around. could be we even want to include Accept() for the comment as well

PR #37, opened 2022-10-08
remove duplicated logging interface, remove log package go module

inconshreveable:
The log package was defined as a separate Go module for reasons that seem unncessary. Technically it would let you implement a Logger without needing to import all of the root module's dependencies, but:

1. If you're working with this library, the end application will almost certainly import the root module anyways.
2. The implementing package can still implement the interface without importing any of our package anyways.

The benefit to this change is:

1. The dependency story is simpler to reason about
2. There is no confusing duplication of a Logger interface in both the main package and the subpackage.

Explicitly pulling @euank as a reviewer since this more or less reverts his earlier PR.

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -4,7 +4,6 @@ go 1.18
 
 require (
 	github.com/ngrok/ngrok-go v0.0.0
-	github.com/ngrok/ngrok-go/log v0.0.0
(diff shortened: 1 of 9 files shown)


euank (approved):
Yup, this is the right fix, thanks!

The main goal was to avoid unneeded dependencies, and this still solves this, I think I just missed that the `*adapter` modules were already each their own go modules last time.

PR #38, opened 2022-10-08
move connect options to functional option pattern

inconshreveable:


jrobsonchase (approved):
Lgtm! Is ConnectAndStartTunnel still around in addition to the new StartTunnel? Might want to kill the longer one and just let people go through the full Connect and then StartTunnel routine. Don't need shorthand for every variation.

inconshreveable:
> Lgtm! Is ConnectAndStartTunnel still around in addition to the new StartTunnel? Might want to kill the longer one and just let people go through the full Connect and then StartTunnel routine. Don't need shorthand for every variation.

ConnectAndStartTunnel is removed in favor of just StartTunnel. now that we have a Session() property on the tunnel object, you can still get access to it from just StartTunnel if you need it

examples/http/main.go
```diff
@@ -21,9 +21,8 @@ func main() {
 
 func run(ctx context.Context) error {
 	_, tun, err := ngrok.ConnectAndStartTunnel(ctx,
```
jrobsonchase, line 23: So does this still compile? I think CI is first on my agenda for next week 😁

jrobsonchase:
Ah, I see. Stupid diff was hidden behind the "file deleted" in the list of changes.

PR #39, opened 2022-10-08
replace AsHTTP with WithHTTPHandler/WithHTTPServer

inconshreveable:
In general, I think this change is for the better.

1. It leads to a smaller interface (removes HTTPTunnel, AsHTTP)
2. It's more type safe, you can only use the HTTP serving functionality on tunnel/endpoint types we know can support it
3. Even if you don't want to use WithHTTPHandler because it's async and doesn't block, it's cleaner to just call http.Serve() on the returned tunnel listener.
4. If we still dislike WithHTTPHandler's async goroutine spawning, I think we just remove it from this PR and still merge to remove HTTPTunnel/AsHTTP. That being said, I think this is still some nice optional sugar for those who want it.


session.go
```diff
@@ -479,10 +480,23 @@ func (s *sessionImpl) StartTunnel(ctx context.Context, cfg config.Tunnel) (Tunne
...
 		Sess:   s,
 		Tunnel: tunnel,
-	}, nil
+	}
+
+	if httpServerCfg, ok := cfg.(interface {
+		HTTPServer() *http.Server
+	}); ok {
+		if srv := httpServerCfg.HTTPServer(); srv != nil {
+			srv.BaseContext = func(_ net.Listener) context.Context { return ctx }
```
jrobsonchase, line 492: I think this might be the first time we're actually using the context passed to StartTunnel for anything *real*. The underlying libraries don't actually use it, so it's just been a future-proofing thing so far.

As a result, we haven't really discussed its semantics. Is it *just* for the Bind rpc call? Is it tied to the lifetime of the tunnel itself, and is canceling it more or less equivalent to calling Close on the tunnel? This change implies the later, but I feel like the former might make more sense.

jrobsonchase (commented):
I'm definitely in favor of removing AsHTTP in favor of using http.Serve like you've changed the examples to, but I'm still pretty on the fence about the tunnel config option, especially wrt Context semantics.

session.go
inconshreveable: you are 100% right. the context should be for the lifetime of the RPC call and not used here. if we had a context for the lifetime of the tunnel itself that makes sense, but in the absence of that it seems like we should just let you use WithHTTPServer and manage that context yourself if you want to supply it

inconshreveable: fixed

inconshreveable:
i've adjusted context semantics. overall i think this is cleaner. ultimately the WithHTTPHandler sugar is optional. i'm okay removing it if we think there's no net benefit but it seems like reasonable sugar that doesn't complicate or detract from the API in any other meaningful ways

inconshreveable:
gentle ping @euank

euank:
I'm on the fence here too.

This feels cleaner in that it's more declarative feeling, which is good! People want to declare up front the tunnels they want IMO, and not deal with the imperative `sess.StartTunnel` API unless they're doing something off the beaten path.

In that regard, this API feels better.

However, the hidden goroutine feels _weird_.

Let's compare the old and new use:


Old:

```go
	tun, err := ngrok.StartTunnel(ctx,
		config.HTTPEndpoint(),
		ngrok.WithAuthtokenFromEnv(),
	)
	err := tun.AsHTTP().Serve(ctx, http.HandlerFunc(handler))
        // http.Server.Serve error
        tun.CloseWithContext(ctx) // possibly a ctx error, so this is wrong :(
        return err
```

New:

```go
	tun, err := ngrok.StartTunnel(ctx,
		config.HTTPEndpoint(
			config.WithHTTPHandler(http.HandlerFunc(handler)),
		),
		ngrok.WithAuthtokenFromEnv(),
	)
	// how do I know if `http.Server.Serve` encountered an error?
	<-ctx.Done()
        tun.CloseWithContext(context.TODO()) // ctx is done, so I can't use it here :(
        return ctx.Err()
```

In practice, `http.Server.Serve` won't error out in many cases. If muxado internally hits a queue overflow or something, I think that bubbles up from `ln.Accept()` and will cause `Serve` to silently stop, but I think that's pretty much the only case, and I can't make such an error happen in practice.

Still, I can't help but want something like:

```go
   select {
   <-tun.Done():
   <-ctx.Done():
    }
    if tun.Err() != nil {
      // err is the underlying `session` error, which propagates to all tunnels on it when the tunnels tear down.
      // In this case, it might be the http.Server.Serve error we're swallowing right now, or some other error like 'session.Stop() called', or 'ctx.Err()', or "the ngrok authtoken was deleted" or something, depending on what happened
    }
```

Having something like that seems compatible with this API, and like it maybe makes it feel strictly better to me?

I do think, on net, the increase in declarativeness is good and I like this more than what we've got now.

jrobsonchase:
Since the implicit use of the `context.Context` used to start the tunnel is no longer a thing, and users are expected to set the `BaseContext` field of their `http.Server`, we can't even use that to block on server exit.

What if the `WithHTTP(Handler|Server)` functions took a channel to return the error through?

```go
func WithHTTPHandler(h http.Handler, errCh chan<- error)
func WithHTTPServer(s *http.Server, errCh chan<- error)
```

Then you have both a way to block on http server exit and a way to get back the error it encountered.

euank:
I think passing in a channel like that is fairly awkward.

It feels more awkward in my eyes than:

```
err := <-tun.ServeErr()
// or
<-tun.Done();
err := tun.Err()
```

inconshreveable:
I think not all sugar needs to be fully functional. Sugar has tradeoffs and not being able to block or pull the error out seems like a pretty reasonable one to me in order to enable a quick way to get an http handler listening. Especially because the alternative is as easy as calling `http.Serve` yourself.

If we really want to take this further I'd suggest a follow up that adds 1) a guarantee that the http server in this option continues to run for the lifetime of the session (by adding a loop into that goroutine) and 2) an API to `Session` that allows the caller to wait on its completion

euank (approved):
Sorry for taking so long to come to an Approve on this :sweat_smile: 

I _think_ this is a net improvement, and you can still plumb in an http server directly which gives you _most_ of the knobs you may want.

I still think there's one thing missing, which is either a "wait on session to finish" function (i.e. `sess.Wait(ctx)` or `<-sess.Done()`), but I think the use of that is a bit of an edge-case, and I'm okay iterating on that separately. It's useful with or without this PR.

I'm okay merging this and iterating further from there.

I do think there might also still be one round left in making a declarative builder API that feels right, but this does seem directionally good.

PR #40, opened 2022-10-11
ci: add initial workflow from the starter example

### LINKED ISSUE #23: Add CI
This should:

- [ ] test
- [ ] lint/fmt
- [ ] tag/release?
- [ ] look for semver regressions?

jrobsonchase:
Tests probably won't pass, since there's no authtoken. I'll address that before mergetime.

Resolves #23 

Does this work?
Resolves ngrok-private/ngrok#13368


jrobsonchase:
Alright, pretty happy with the test groups for the time being. We get a few basic unauthenticated tests to run at the moment, as well as any fully-offline tests.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,43 @@
...
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.50.0
+
+          # Optional: working directory, useful for monorepos
+          # working-directory: somedir
+
+          # Optional: golangci-lint command line arguments.
+          args: --skip-dirs=internal/muxado
```
euank, line 30: I think we should remove this from `lint.yml` and put it in a `.golangci.yml` in the root with:

```yaml

run:
  skip-dirs:
  - internal/muxado
```

The intent is of course that a naive runner of just `golangci-lint` doesn't get warnings or such.

euank (commented):
One suggestion, otherwise looks good!

euank (approved):
lgtm!

PR #42, opened 2022-10-14
loggers: Remove pgx, add zap and logrus, remove 'adapter' from module names

jrobsonchase:


log/log15/adapter.go
```diff
@@ -1,10 +1,26 @@
-package log15adapter
+// Package logrus provides a logger that writes to a
```
inconshreveable, line 1: logrus -> log15

inconshreveable (approved):
LGTM

PR #43, opened 2022-10-14
Add OIDC support

jrobsonchase:
Also clean up oauth a bit, and update the go.work file.


config/oidc.go
```diff
@@ -0,0 +1,68 @@
...
+		for _, opt := range opts {
+			opt(oidc)
+		}
+
+		cfg.OIDC = oidc
+	})
+}
+
+// Append email addresses to the list of allowed emails.
+func WithOIDCEmail(addr ...string) OIDCOption {
```
inconshreveable, line 50: should this (and the ones for oauth too) be `WithOIDCAllowEmail`, `WithOAuthAllowDomain`, etc? i think so

config/oidc.go
```diff
@@ -0,0 +1,68 @@
...
+		ClientSecret: oidc.ClientSecret,
+		AllowEmails:  oidc.AllowEmails,
+		AllowDomains: oidc.AllowDomains,
+		Scopes:       oidc.Scopes,
+	}
+}
+
+// WithOIDC configures this edge with the the given OIDC provider.
+// Overwrites any previously-set OIDC configuration.
+func WithOIDC(issuerURL string, clientID string, clientSecret string, opts ...OIDCOption) HTTPEndpointOption {
```
inconshreveable, line 33: is three required parameters enough to break it out into a struct? probably not, but maybe?

inconshreveable (approved):

config/oidc.go
jrobsonchase: Yeah, I think it's *just* few enough to avoid a struct. We've managed to avoid any extra structs in our config interface with the functional options, so I think it makes sense to err on the side of "no structs" unless it's absolutely awful without them.

PR #44, opened 2022-10-14
proto: add ClientType and declare this library as a library

jrobsonchase:
Also sets the library version back to 0.0.0. This means we'll have to wait on ngrok-private/ngrok#13827 to merge/deploy, or else it'll get rejected for having too low a version again.

### DIFF
--- a/internal/tunnel/proto/msg.go
+++ b/internal/tunnel/proto/msg.go
@@ -72,7 +72,16 @@ type AuthExtra struct {
 	ConfigVersion   string
 	CustomInterface bool
 	CustomCAs       bool
+
+	ClientType ClientType // The type of client this is. Currently agent and library clients are supported
 }
+
+type ClientType string
+
+const (
+	Agent   ClientType = "agent"
+	Library ClientType = "library"
+)
 
 type Fingerprint struct {
 	M []string
--- a/session.go
+++ b/session.go
@@ -351,7 +351,7 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
 		StopUnsupportedError:    remoteStopErr,
 		UpdateUnsupportedError:  remoteUpdateErr,
 
-		// TODO: More fields here?
+		ClientType: proto.Library,
 	}
 
 	reconnect := func(sess tunnel_client.Session) error {
--- a/session.go
+++ b/session.go
@@ -26,7 +26,7 @@ import (
 )
 
 // The ngrok library version.
-const libraryAgentVersion = "3.0.0-ngrok-go"
+const libraryAgentVersion = "0.0.0"
 
 // The interface implemented by an ngrok session object.
 type Session interface {


session.go
```diff
@@ -26,7 +26,7 @@ import (
 )
 
 // The ngrok library version.
-const libraryAgentVersion = "3.0.0-ngrok-go"
+const libraryAgentVersion = "0.0.0"
```
euank, line 29: Should we leave in `+golang` still? Seems like something that'll be handy to have in the version, unless there's a more obvious place it is session-level?

euank (approved):
:+1:

session.go
jrobsonchase: Yeah, the version isn't the place to put it. Semver-wise, it's a version tag that just marks it as "incompatible" iirc.

Another field like "library id" would be the more correct approach. I even had it over in #13827 before deciding that I didn't really have a use for it yet, and so omitted it.

euank: The semver spec calls the stuff after the `+` "Build metadata", which doesn't seem that wrong of a place to put the language.

I agree having more properly structured fields would be better. It does seem like something worth having before a general release though since I think we'll certainly want metrics on language client usage, and also want that to be visible in metrics for identifying any issues with specific clients/versions.

internal/tunnel/proto/msg.go
```diff
@@ -72,8 +72,17 @@ type AuthExtra struct {
 	ConfigVersion   string
 	CustomInterface bool
 	CustomCAs       bool
+
+	ClientType ClientType // The type of client this is. Currently agent and library clients are supported
```
inconshreveable, line 76: instead of Agent | Library, should this be Agent | GolangLibrary | JavaLibrary | etc?

jrobsonchase: Mmmm, maybe?

Realistically, I think we'll have a different protocol altogether before we get may more libraries, so just agent/library is probably fine for right now.

inconshreveable: i think it's going to be better long term to just treat this as a "User Agent" where each language is different so that it can be unique across each of them. it's not any extra handling server side and it gives us more fidelity

jrobsonchase: OK, I'm packing more info into it over in ngrok-private/ngrok#13848. Any thoughts on the information/structure there?

euank: The `official` thing seems kinda.. I dunno, redundant?
Would it ever _not_ be marked official?

jrobsonchase: Not by us, but since the library is open-source, nothing stops people from forking it and creating unofficial ngrok libraries.

We also can't stop them from simply continuing to claim that they're official, but my thinking is that this at least gives them the opportunity to play nice.

PR #45, opened 2022-10-28
Update import path from github -> golang.ngrok.com

euank:
This allows us to have `ngrok-go` in the github URL, without having the slightly confusing import basename vs package name mismatch.

As-is, this change depends on an internal change for that URL to go public, but with some local dns hackery, I can see it work via:

```
$ GOPRIVATE=golang.ngrok.com GOPROXY=direct GOINSECURE=golang.ngrok.com go get -v -u golang.ngrok.com/ngrok@euan/update-import-path
```

being able to succeed (depending on me having setup my local dns to point `golang.ngrok.com` at an internal testing domain)

### DIFF
--- a/internal/pb_agent/middleware.pb.go
+++ b/internal/pb_agent/middleware.pb.go
@@ -7,10 +7,11 @@
 package pb_agent
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
+	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
+	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 )
 
 const (
--- a/config/basic_auth.go
+++ b/config/basic_auth.go
@@ -1,6 +1,6 @@
 package config
 
-import "github.com/ngrok/ngrok-go/internal/pb_agent"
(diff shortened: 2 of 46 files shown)


euank:
This will, of course, break all our users if they `go get -u` the module after we merge this, but during alpha is the time to do that :)

internal/pb_agent/middleware.pb.go
```diff
@@ -7,10 +7,11 @@
 package pb_agent
 
 import (
-	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
-	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
 	reflect "reflect"
 	sync "sync"
+
+	protoreflect "google.golang.org/protobuf/reflect/protoreflect"
+	protoimpl "google.golang.org/protobuf/runtime/protoimpl"
```
jrobsonchase, line 14: This is an odd change - is our lint test not catching non-gofmt'd changes?

euank: This was correctly gofmt'd before, but not goimport'd.

We can update the lint separately to enforce that if we want to have something more strict than just gofmt.
I just ran goimports to group the new import path more consistently, and it happened to also pick this up.

euank:
Verified that:

```
GOPROXY=direct go get -u golang.ngrok.com/ngrok@euan/update-import-path
```

works with this commit now, since dns is now live. I think this'll all work without `GOPROXY=direct` post-merge now too!

PR #46, opened 2022-11-04
config: test generated protocol requests

jrobsonchase:
Add a whole mess of tests that verify that all of our functional options actually produce the correct bind requests.

Tested with code coverage to get reasonably sure that we're hitting all of the code paths.
[cover.out.txt](https://github.com/ngrok/ngrok-go/files/9940311/cover.out.txt)

PR #50, opened 2022-11-15
proto: deprecate/remove hostname and subdomain

jrobsonchase:
Follow-up from #46.

Removes the deprecated `Hostname` and `Subdomain` fields for the new and improved `Domain` field.

No externally-visible API changes.

### DIFF
--- a/online_test.go
+++ b/online_test.go
@@ -400,27 +400,6 @@ func TestProxyProto(t *testing.T) {
 	}
 }
 
-func TestHostname(t *testing.T) {
-	paidTest(t)
-	ctx := context.Background()
-
-	tun, exited := serveHTTP(ctx, t, nil,
-		config.HTTPEndpoint(config.WithDomain("foo.robsonchase.com")),
-		helloHandler,
-	)
-	require.Equal(t, "https://foo.robsonchase.com", tun.URL())
-
-	resp, err := http.Get(tun.URL())
-	require.NoError(t, err)
-
-	content, err := io.ReadAll(resp.Body)
-	require.NoError(t, err)
-	require.Equal(t, "Hello, world!\n", string(content))
-
-	require.NoError(t, tun.CloseWithContext(ctx))
-	require.Error(t, <-exited)
-}
-
 func TestSubdomain(t *testing.T) {
 	paidTest(t)
 	ctx := context.Background()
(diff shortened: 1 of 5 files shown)


online_test.go
```diff
@@ -400,15 +400,29 @@ func TestProxyProto(t *testing.T) {
...
+	prefix := randString(8)
+	domain := fmt.Sprintf("%s.ngrok.io", prefix)
+	expectedUrl := fmt.Sprintf("https://%s", domain)
+
 	tun, exited := serveHTTP(ctx, t, nil,
-		config.HTTPEndpoint(config.WithDomain("foo.robsonchase.com")),
+		config.HTTPEndpoint(config.WithDomain(domain)),
 		helloHandler,
 	)
-	require.Equal(t, "https://foo.robsonchase.com", tun.URL())
```
jrobsonchase, line 411: This probably wouldn't have worked for anyone but me, or someone authenticated with our dev account. Whoops :stuck_out_tongue:

nikolay-ngrok: and still might fail ;) How does randomly server-selected domain work? Just don't pass `WithDomain` option?

online_test.go
```diff
@@ -400,15 +400,29 @@ func TestProxyProto(t *testing.T) {
...
 	paidTest(t)
 	ctx := context.Background()
 
+	prefix := randString(8)
+	domain := fmt.Sprintf("%s.ngrok.io", prefix)
+	expectedUrl := fmt.Sprintf("https://%s", domain)
+
 	tun, exited := serveHTTP(ctx, t, nil,
-		config.HTTPEndpoint(config.WithDomain("foo.robsonchase.com")),
+		config.HTTPEndpoint(config.WithDomain(domain)),
```
nikolay-ngrok, line 422: Does this change in any ways `TestSubdomain`? Cause it looks kind of the same? Also `ngrok.io`?

online_test.go
jrobsonchase: Yeah, but this test is specifically to test the option, so falling back to a random one wouldn't really make sense :stuck_out_tongue:

online_test.go
jrobsonchase: *Huh*, they are the same! :sweat_smile: 

Cool, just gonna axe the one I was messing with then.

nikolay-ngrok (approved):
👍

PR #52, opened 2022-11-28
internalize log15

jrobsonchase:
See https://github.com/ngrok-private/ngrok/issues/14472.

I don't really like this solution all that much, but it *technically* works.

If we can cut a new release of log15 that makes the problem go away, I'd much prefer that.


inconshreveable:
let's just cut a new release of log15 then?

jrobsonchase:
Superseded by #55

PR #56, opened 2022-11-29
proto: pull in changes from our internal version

jrobsonchase:
Most notably, this includes the change from the go-specific generated-protobuf-json representation to a more portable wire-format byte blob.


internal/tunnel/proto/msg.go
```diff
@@ -79,13 +82,12 @@ type AuthExtra struct {
...
-	Agent   ClientType = "agent"
-	Library ClientType = "library"
+	LibraryOfficialGo ClientType = "library/official/go"
 )
 
 type Fingerprint struct {
-	M []string
-	D []string
+	M []string // mac addresses
+	D []string // disk serial numbers
```
euank, line 90: I manually stripped those before since, well, they're unused in this library
(we also don't actually populate D in the agent, and iirc never did, and I'd rather not give the impression we did)

euank (approved):

PR #58, opened 2022-12-01
ci: use direnv to ensure consistent environment

euank:
Fixes https://github.com/ngrok/ngrok-go/issues/57

`apt install` was getting us an old protoc version that didn't handle go
modules correctly. This fixes that by using direnv to ensure it's the
same for all of us.

Unfortunately, `apt install` also installed an old direnv and doesn't
have nix-direnv, so I added some more dependencies to install a newer
direnv.

Also of note, I merged fmt/gen/lint into one so we only have to install
nix once. I think that the savings of running them in parallel are
minimal compared to the time it takes to install nix/direnv


jrobsonchase (approved):
Nice! I'd been putting off figuring out how to get better CI/local dev parity.

PR #71, opened 2023-03-03
oauth: support client id/secret configuration

### LINKED ISSUE #70: Expose OAuth ClientID/ClientSecret configuration
Seems these were missed in the first pass. Shouldn't be too hard to add.

jrobsonchase:
Resolves #70

Also adding a hack to the generated protobuf file, since I don't really
want to figure out how to make protoc do it for me.


jrobsonchase:
Ahh, crap. Our linter doesn't like my protobuf hack :disappointed:

jrobsonchase:
OK, reverted the obfuscation in the protoc-generated types. Guess I'll save that for another day.

It's at least no worse than it was before.

natasha-jarus (approved):
nice!

PR #75, opened 2023-03-08
Comments refactor for docs

thebluetiger:


config/tls.go
```diff
@@ -18,7 +18,7 @@ func (of tlsOptionFunc) ApplyTLS(cfg *tlsOptions) {
 	of(cfg)
 }
 
-// Construct a new set of HTTP tunnel options.
+// Construct a new set of TLS tunnel options.
```
jrobsonchase, line 21: While you're here, we should also start comments with the name of the item being documented, e.g.

```
// TLSEndpoint constructs a new set of options for a TLS endpoint.
```

jrobsonchase (approved):

PR #78, opened 2023-03-14
Add handlers for Restart and Update server commands, plus fancy unimplemented handler errors

natasha-jarus:
Expose a few more features of the underlying session protocol. Offering `Restart` alongside `Stop` seems reasonable; `Update` is a bit more odd but still a neat feature to offer for more complex applications.

Internally, the `Update` message has a couple fields that could be exposed; however, ngrok internally doesn't use them and they're even less useful in the context of third-party applications, so I don't think there's a need to make those fields available.

Allowing application authors to set reasons for disabling server command handlers is useful in cases where those handlers are only supported on particular platforms or under specific circumstances. Does the documentation for those options make it sufficiently clear that not setting either a handler or a command disabled option still results in a disabled server command handler with a default "unimplemented" error reported in the dashboard/API?

### DIFF
--- a/session.go
+++ b/session.go
@@ -115,6 +115,10 @@ type connectConfig struct {
 	RestartHandler ServerCommandHandler
 	UpdateHandler  ServerCommandHandler
 
+	remoteStopErr    *string
+	remoteRestartErr *string
+	remoteUpdateErr  *string
+
 	// The logger for the session to use.
 	Logger log.Logger
 }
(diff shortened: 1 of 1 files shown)


jrobsonchase (approved):
Seems reasonable to me 👍 

I think we originally axed the Restart/Update handlers because we didn't want to answer the "what should these actually *do*?" question, but we're exposing them in Rust and they're needed for the agent to do its thang, so might as well 🤷 

A bit less sure about letting users customize the "not implemented" error strings though. Does the agent *really* need that functionality to surface meaningful information, or are we just trying to avoid changing the strings unnecessarily?

natasha-jarus:
my two cents is: as an end user, "the agent has not defined a callback for this command" is a lot more confusing than "you can't remote restart windows agents" or the like. i reckon most folks seeing those strings are not going to be the application authors, so a little polish is worth the effort.

PR #79, opened 2023-03-14
session: add user-agent string and allow library consumers to extend it

natasha-jarus:
Allows library consumers to differentiate their application and version from others.

Currently, there is no incentive for folks to use this outside of caring about our internal usage; we should expose this to users via the dashboard & API in the future.

### DIFF
--- a/session.go
+++ b/session.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"regexp"
 	"runtime"
 	"strings"
 	"sync/atomic"
@@ -75,6 +76,37 @@ type ServerCommandHandler func(ctx context.Context, sess Session) error
 // establishment.
 type ConnectOption func(*connectConfig)
 
+type clientInfo struct {
+	Type    string
+	Version string
+}
+
+var bannedUAchar = regexp.MustCompile("[^!#$%&'*+-.^_`|~0-9a-zA-Z]")
+
+// Formats client info as a well-formed user agent string
+func (c *clientInfo) ToUserAgent() string {
+	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version)
+}
+
+func sanitizeUserAgentString(s string) string {
+	s = strings.ReplaceAll(s, "/", "-")
+	s = bannedUAchar.ReplaceAllString(s, "#")
+	return s
+}
+
+// version, type, user-agent
+func generateInfo(cs []clientInfo) (string, string, string) {
+	var versions, types, uas []string
+
+	for _, c := range cs {
+		versions = append(versions, c.Version)
+		types = append(types, c.Type)
+		uas = append(uas, c.ToUserAgent())
+	}
+
+	return strings.Join(versions, ","), strings.Join(types, ","), strings.Join(uas, " ")
+}
+
 // Options to use when establishing an ngrok session.
 type connectConfig struct {
 	// Your ngrok Authtoken.
@@ -99,6 +131,10 @@ type connectConfig struct {
 	// Opaque metadata string to be associated with the session.
 	// Viewable from the ngrok dashboard or API.
 	Metadata string
+
+	// Child client types and versions used to identify specific applications
+	// using this library to the ngrok service.
+	ClientInfo []clientInfo
 
 	// HeartbeatInterval determines how often we send application level
 	// heartbeats to the server go check connection liveness.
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -133,6 +138,20 @@ func WithMetadata(meta string) ConnectOption {
 	}
 }
 
+// WithUserAgent configures the user-agent string for this session.
+// This is a means for you to identify your application to ngrok and to
+// differentiate it from other applications using the agent library.
+//
+// We recommend to set this with your application name and version following
+// the format specified in RFC 7231. See [MDN's user-agent documentation] for examples.
```
euank, line 146: I'm not 100% sure on the recommendation here. Currently, this doesn't show up in the dashboard at all, right, so this is really just currently useful for usage metrics.

For usage metrics, we'd only want unique user agents for quite large buckets (so i.e. a library using this and expecting other users of it), but not for individual applications (like a once-off webhook test, if I set that to 'my-unpublished-testing-app', that actually gives less information than the default 'ngrok-go' one).

I actually kinda think the right recommendation might be "This value should typically be left unset. If you are creating a library or application you intend to publish broadly, setting a UserAgent may make sense. (etc etc RC 7231)"

.... Of course, the above is based on UserAgent being the unpublished internal-only thing.

I could also see us changing it so "UserAgent" is exposed on the session and we encourage changing it, and "Transport" gets updated to match what UserAgent is now (i.e. `ngrok/2` for ngrok agents, `ngrok-go/0` for ngrok-go, etc etc), and that would then give us information on library usage, while still letting people tweak user agents.

jrobsonchase (changes requested):
For now, could we mirror the [rust implementation of this](https://github.com/ngrok/ngrok-rs/pull/70/files)?

I expect that this will need to evolve as soon as we get the minimum agent version spec figured out, but we might as well use the same temporary approach for the time being 😛

natasha-jarus:
oh, yeah, i can do that. for what it's worth, the agent currently publishes its user-agent as `ngrok/3.2.1` and part of my reasoning for following the UA RFC was so that we'd keep that string construct for whenever we get around to parsing user agents if we want nicer bucketing.

session.go
natasha-jarus: i think eventually people will want to be able to see application names in the dashboard and API so they can quickly see which things are run by which app without having to divine this by correlating IPs or URLs or whatever. what that looks like is very TBD!

session.go
```diff
@@ -398,10 +420,19 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
...
+	cfg.ClientTypes = append([]string{"library/official/go"}, cfg.ClientTypes...)
+	cfg.ClientVersions = append([]string{libraryAgentVersion}, cfg.ClientVersions...)
+
+	userAgent := fmt.Sprintf("%s %s",
+		strings.Join(cfg.ClientTypes, ","),
+		strings.Join(cfg.ClientVersions, ","),
+	)
+
 	auth := proto.AuthExtra{
 		Version:            libraryAgentVersion,
```
euank, line 432: Shouldn't this now be:

```go
Version: strings.Join(cfg.ClientVersions, ","),
....
UserAgent: strings.Join(cfg.ClientTypes, ","),
```

If I'm reading the rust code right, something like `child_client("foo", "1.0.0").child_client("bar", "2.0.0")` results in `UserAgent: "foo, bar", Version: "1.0.0,2.0.0"`, and I guess we have to zip those two lists together to correlated later.

If we want to mimic that, I think we want them split here too, right?

natasha-jarus: apparently i cannot read, but also that's an interesting departure from the way we've used User-Agent up till now; it's always had a version string before. Both?

euank (approved):
lgtm

session.go
```diff
@@ -398,10 +420,17 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
...
+
+	version := strings.Join(cfg.ClientVersions, ",")
+	userAgent := strings.Join(cfg.ClientTypes, ",")
+
 	auth := proto.AuthExtra{
-		Version:            libraryAgentVersion,
+		Version:            version,
 		Authtoken:          proto.ObfuscatedString(cfg.Authtoken),
 		Metadata:           cfg.Metadata,
+		UserAgent:          userAgent,
```
jrobsonchase, line 433: This still needs to be the `ClientType` field (down on line 443) instead of the `UserAgent` field to match Rust.

Edit: Wait, we already had a `UserAgent` field? And `ClientType` is actually the newer field?? Why didn't I just use that in the first place? Someone should have stopped me! 😅

jrobsonchase (approved):
Ugh, I'm an idiot. 🤦

For now, let's just merge this (after the UserAgent -> ClientType change), and I'll deal with deduplicating the approach later.

session.go
natasha-jarus: Wait there's a `ClientType` field? :rofl:

okay i'm going to set that AND set the user agent; opinions on whether the UA follows `name1/version1 name2/version2` or `name1,name2/version1,version2` format?

session.go
```diff
@@ -75,6 +76,37 @@ type ServerCommandHandler func(ctx context.Context, sess Session) error
...
+	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version)
+}
+
+func sanitizeUserAgentString(s string) string {
+	s = strings.ReplaceAll(s, "/", "-")
+	s = bannedUAchar.ReplaceAllString(s, "#")
+	return s
+}
+
+// version, type, user-agent
```
natasha-jarus, line 97: don't love this, also don't want to splat a dozen single-use structs around :(

jrobsonchase (approved):
Looks good!

Only change I'd make is to un-export that ClientConfig type.

session.go
```diff
@@ -75,6 +76,37 @@ type ServerCommandHandler func(ctx context.Context, sess Session) error
 // establishment.
 type ConnectOption func(*connectConfig)
 
+type ClientInfo struct {
```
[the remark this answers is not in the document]
natasha-jarus: done!

PR #80, opened 2023-03-17
session: allow ReconnectingSession to retry temporary failures

natasha-jarus:
In the case of a temporary failure in ReconnectingSession, an error is placed into the stateChanges channel. This error does not necessarily mean that failure is permanent and that the connection process should abort. For example, in the case of a dial error when initially connecting the session, we desire to retry dialing with backoff as implemented by ReconnectingSession, unless the library consumer intentionally wants to cancel connecting based on that error.

Only in the case of a permanent connection failure do we want to cancel the connection process and return an error to the consumer.

Therefore, we expose all connection errors to the library consumer via the Disconnect handler, allowing them, if they so choose, to .Close() an open session that is erroring (e.g., on authentication or restarting tunnel binds). The Connect function returns only upon permanent failure, in which case it returns nil and the last error read from stateChanges; or upon connection success, in which case it returns a happy Session. In the case of permanent failure, the Disconnect handler will have already received the error; thus, for consumers defining a Disconnect handler it is not necessary to handle the error returned from Connect.

This PR also ensures the guarantee that whenever the underlying session closes, the Disconnect handler is called with `err = nil`.

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -1,10 +1,8 @@
 package ngrok
 
 import (
-	"errors"
 	"fmt"
 	"net/url"
-	"strings"
 )
 
 // Errors arising from authentication failure.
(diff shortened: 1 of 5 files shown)


session.go
```diff
@@ -551,41 +551,61 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
...
+	// allow consumers to .Close() the session before a successful connect
+	session.setInner(&sessionInner{
+		Session: sess,
+	})
+
+	// performs one "pump" of the session update channel
+	// returns true if there are more updates to handle
+	runSessionHandlers := func() (bool, error) {
+		select {
+		case <-ctx.Done():
```
natasha-jarus, line 563: i think we might want to run the disconnect handler here too, since otherwise consumers have no way of knowing their session died due to a done context after the initial connect succeeds.

errors.go
```diff
@@ -113,3 +115,56 @@ func (e errSessionDial) Is(target error) bool {
 	_, ok := target.(errSessionDial)
 	return ok
 }
+
+type errMultiple struct {
+	inners []error
+}
```
jrobsonchase, line 121: Don't we already pull in multierr (or similar)? Any reason we can't simply use that?

Edit: Oh, apparently not. Well, might consider using it anyway? Or maybe we'd prefer avoiding dependencies ¯\\\_(ツ)\_/¯

jrobsonchase (approved):
looks good to me!

jrobsonchase (approved):

PR #82, opened 2023-03-27
add connection proxy metadata

natasha-jarus:
Expose the tunnel proto, edge type, and passthrough tls for each connection to a tunnel. Re-export the internal edge type constants so consumers can use constants instead of magic number strings.


jrobsonchase (approved):
Looks good to me!

PR #84, opened 2023-04-03
session: allow custom SNIs for very rare situations

natasha-jarus:
In 100% of customer use cases the SNI set by ngrok-go is correct. However, in the case that someone decides to, say, hard-code an IP address for the tunnel ingress (or otherwise do the DNS lookup for it before calling the agent), they may want to set an appropriate SNI so the server can validate the tls cert.

We use this feature internally as part of our end to end test suite. :)

### DIFF
--- a/session.go
+++ b/session.go
@@ -114,6 +114,9 @@ type connectConfig struct {
 	// The address of the ngrok server to connect to.
 	// Defaults to `tunnel.ngrok.com:443`
 	ServerAddr string
+	// The SNI used to connect to the ngrok server.
+	// Defaults to the hostname in [ServerAddr]
+	SNI string
 	// The [x509.CertPool] used to authenticate the ngrok server certificate.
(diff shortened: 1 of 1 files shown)


euank (approved):
:+1:

euank:
I'll comment that there might be a better change instead: `WithTLSConfig(*tls.Config)` (or `WithTLSDialer`)

Someone might wish to override more parts of the TLS config, and it seems quite reasonable to allow that too for similar unusual cases.

jrobsonchase (approved):
LGTM as well :+1: 

Agreed with @euank that being able to provide a full tls config would probably be useful as well - that's what we do over in Rust land. It lets library consumers make whatever tweaks are needed, and then if we find ourselves repeating a pattern over and over, we add shortcuts.

natasha-jarus:
superseded by #85

PR #85, opened 2023-04-04
session: allow custom TLS configuration

natasha-jarus:
Allow library consumers to customize the TLS configuration used by the agent on connection. We expose this as a callback, rather than a direct override of the TLS config, so that users can make tweaks without having to e.g. hardcode a region server name or the like.

### DIFF
--- a/session.go
+++ b/session.go
@@ -114,6 +114,8 @@ type connectConfig struct {
 	// The address of the ngrok server to connect to.
 	// Defaults to `tunnel.ngrok.com:443`
 	ServerAddr string
+	// The [tls.Config] used when connecting to the ngrok server
+	TLSConfigCustomizer func(*tls.Config)
 	// The [x509.CertPool] used to authenticate the ngrok server certificate.
 	CAPool *x509.CertPool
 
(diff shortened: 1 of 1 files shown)


bobzilladev (approved):
Looks reasonable, and like josh said is analogous to an option on the rust side. Would be nice to see a use of it working, either as a test case here, or a link to it being used by the agent for SNI setting or otherwise.

session.go
```diff
@@ -260,6 +262,15 @@ func WithServer(addr string) ConnectOption {
 	}
 }
 
+// WithTLSConfig allows customization of the TLS connection made from the agent
+// to the ngrok service. Customization is applied after the [WithServer] and
+// [WithCA] options are applied.
```
euank, line 267: We could also mention that "WithCA" (and I think "WithServer" too, right?) have no impact if this is set.

euank (approved):
:+1:

jrobsonchase (commented):
I'm not 100% in love with the callback approach, but I do see the benefits. It just introduces more complexity/boilerplate for the caller. 

I think the only change I'd make is to omit the returned value and just let the callback modify its argument. That's how people are going to do it anyway.

session.go
natasha-jarus: yeah, `WithCA` and `WithServer` (and `WithRegion`) do affect what gets put in the "base" tls config!

jrobsonchase (approved):

PR #88, opened 2023-04-07
Deprecate WithTermination in favor of clearer WithEdgeTermination and WithCustomEdgeTermination

### LINKED ISSUE #86: Not clear in documentation how to achieve the same functionality of `ngrok tls --terminate-at edge`
I am working on integrating ngrok in an internal workflow here. Because of particularities of the protocol used, I need it to use with TLS terminating the encryption at the edge aka `ngrok tls --terminate-at edge`.

The Go SDK doesn't document how to achieve that and it seems to default to some other mode.

Appreciate the difference between these two setups:
```
	tun, err := ngrok.Listen(ctx,
		config.TLSEndpoint(
			config.WithDomain(ngrokDomain),
			config.WithTermination([]byte{}, []byte{}),
		),
		ngrok.WithAuthtokenFromEnv(),
	)
```

vs

```
	tun, err := ngrok.Listen(ctx,
		config.TLSEndpoint(
			config.WithDomain(ngrokDomain),
		),
		ngrok.WithAuthtokenFromEnv(),
	)
```

jrobsonchase:
Resolves #86

We'll probably want to add a `WithAgentTermination` as well, but that
will be a little more involved, and may fit into the built-in connection
forwarding better.

### DIFF
--- a/config/tls_termination.go
+++ b/config/tls_termination.go
@@ -1,10 +1,81 @@
 package config
 
+type TLSTerminationLocation int
+
+const (
+	// Terminate TLS at the ngrok edge. The backend will receive a plaintext
+	// stream.
+	TLSAtEdge TLSTerminationLocation = iota
+	// Terminate TLS in the ngrok library. The library will receive the
+	// handshake and perform TLS termination, and the backend will receive the
+	// plaintext stream.
+	// TODO: export this once implmented
+	tlsAtLibrary
+)
+
+type tlsTermination struct {
+	location TLSTerminationLocation
+	key      []byte
+	cert     []byte
+}
+
+func (tt tlsTermination) ApplyTLS(cfg *tlsOptions) {
+	switch tt.location {
+	case tlsAtLibrary:
+		cfg.KeyPEM = nil
+		cfg.CertPEM = nil
+		// TODO: implement this in the tunnel `Accept` call.
+		panic("automatic tls termination in-app is not yet supported")
+	case TLSAtEdge:
+		cfg.KeyPEM = tt.key
+		cfg.CertPEM = tt.cert
+		return
+	}
+}
+
+type TLSTerminationOption func(tt *tlsTermination)
+
+// WithTLSTermination arranges for incoming TLS connections to be automatically terminated.
+// The backend will then receive plaintext streams, rather than raw TLS connections.
+// Defaults to terminating TLS at the ngrok edge with an automatically-provisioned keypair.
+func WithTLSTermination(opts ...TLSTerminationOption) TLSEndpointOption {
+	tt := tlsTermination{
+		location: TLSAtEdge,
+		key:      []byte{},
+		cert:     []byte{},
+	}
+	for _, opt := range opts {
+		opt(&tt)
+	}
+	return tt
+}
+
 // WithTermination sets the key and certificate in PEM format for TLS termination at the ngrok
 // edge.
+//
+// Deprecated: Use WithCustomEdgeTermination instead.
 func WithTermination(certPEM, keyPEM []byte) TLSEndpointOption {
 	return tlsOptionFunc(func(cfg *tlsOptions) {
 		cfg.CertPEM = certPEM
 		cfg.KeyPEM = keyPEM
+	})
+}
+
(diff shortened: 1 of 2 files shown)


jrobsonchase:
* main
  * **PR #88** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-go/88" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-go/88?utm_source=stack-comment).

jrobsonchase:
cc @russorat and/or @inconshreveable for API opinions

russorat (approved):
lgtm

inconshreveable:
don't love it. why not `WithTLSTermination(... TLSTerminationOption)` changing to take two functional options:

- one that disables it (because it's on by default) (e.g. `WithTLSTerminationDisabled`)
- a second that let's the caller specify a cert/key to ship to use for termination (`WithTLSTermationCertKeyPair`)

jrobsonchase:
@inconshreveable I'm not sure I see the benefit of having another layer of functional option wrapping over letting those options simply being top-level TLS options.

also:
> one that disables it (because it's on by default) (e.g. WithTLSTerminationDisabled)

No it's not - by default we pass the TLS handshake down to the agent.

There are three termination scenarios for TLS endpoints:
* No termination - agent receives the TLS stream from the client verbatim. Either the agent or backed has to terminate it. (the default)
* Edge termination with user-provided certificates. This is the only one we advertised support for.
* Edge termination with ngrok-provided certificates. This one was possible "by accident" if the user passes empty slices to the existing `WithTermination` function.

inconshreveable:
Indeed, the ngrok agent's default is not what I expected. even so, it feels like we're configuring a single behavior and these are configuration knobs you can twiddle within that larger behavior. projecting out a fuller api to explain why it should be constructed like this, i'd propose this is a reasonable future state which supports the proposed design:

```
WithTLSTermination(
  WithTLSTerminationEnabled(true|false),
  // we could support this in the library
  WithTLSTerminationAt(Edge|Agent),
  // eventually if we wanted this kind of configurability, these three below could be collapsed to `WithTLSConfig(*tls.Config)`
  WithTLSTerminationCertKeyPair(cert, key []byte),
  WithTLSTerminationMinimumVersion(int),
  WithTLSTerminationCipherSuites(...CipherSuite),
)
```

thoughts?

jrobsonchase:
@inconshreveable Going through my backlog, realized I forgot this was in flight.

Updated to be more in-line with your proposed interface.

config/tls_termination.go
```diff
@@ -1,10 +1,81 @@
...
+	cert     []byte
+}
+
+func (tt tlsTermination) ApplyTLS(cfg *tlsOptions) {
+	switch tt.location {
+	case tlsAtLibrary:
+		cfg.KeyPEM = nil
+		cfg.CertPEM = nil
+		// TODO: implement this in the tunnel `Accept` call.
+		panic("automatic tls termination in-app is not yet supported")
```
inconshreveable, line 28: can we return this as an error to the caller instead of panic'ing? if not, we should probably document it loudly

inconshreveable (approved):

config/tls_termination.go
jrobsonchase: Not in an ergonomic way. You'd have to do something like:

```go
opt, err := config.WithTerminateTLSAt(config.TLSAtEdge)
if err != nil {
  return err
}

tun, err := config.TLSEndpoint(opt)
// etc.
```

rather than calling the functional option inline. Could probably rearchitect it such that the options are run at tunnel bind time, and then we could surface any configuration errors in the listen call, but that doesn't feel worth doing just for this.

I'm OK with it panicing for now since the constant isn't exported. Someone would have to cast an integer like `config.WithTLSTerminationAt(config.TLSTerminationLocation(1))`, and in that case I feel like a panic is well-deserved :grin:

inconshreveable: > Could probably rearchitect it such that the options are run at tunnel bind time, and then we could surface any configuration errors in the listen call, but that doesn't feel worth doing just for this.

we could also just record the error and surface it at listen time? agreed not a high priority

PR #92, opened 2023-04-20
session: add optional comments to user agent string

natasha-jarus:
Just what it says on the tin. RFC 7231/7230 compliance here we come!

### DIFF
--- a/session.go
+++ b/session.go
@@ -78,15 +78,20 @@ type ServerCommandHandler func(ctx context.Context, sess Session) error
 type ConnectOption func(*connectConfig)
 
 type clientInfo struct {
-	Type    string
-	Version string
+	Type     string
+	Version  string
+	Comments []string
 }
 
 var bannedUAchar = regexp.MustCompile("[^!#$%&'*+-.^_`|~0-9a-zA-Z]")
 
 // Formats client info as a well-formed user agent string
 func (c *clientInfo) ToUserAgent() string {
-	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version)
+	comment := ""
+	if len(c.Comments) > 0 {
+		comment = fmt.Sprintf(" (%s)", strings.Join(c.Comments, "; "))
+	}
+	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version) + comment
 }
 
 func sanitizeUserAgentString(s string) string {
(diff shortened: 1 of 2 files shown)


session.go
```diff
@@ -78,15 +78,21 @@ type ServerCommandHandler func(ctx context.Context, sess Session) error
...
 
 // Formats client info as a well-formed user agent string
 func (c *clientInfo) ToUserAgent() string {
-	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version)
+	comments := make([]string, 0, len(c.Comments))
+	for _, comment := range c.Comments {
+		comments = append(comments, fmt.Sprintf(" (%s)", comment))
+	}
+
+	return sanitizeUserAgentString(c.Type) + "/" + sanitizeUserAgentString(c.Version) + strings.Join(comments, " ")
```
jrobsonchase, line 95: You've got an extra space in either your Join or your Sprintf. Probably be easiest to just change your Join to use `""` since you'll want that leading space from the `Sprintf` anyway.

session.go
```diff
@@ -184,9 +190,9 @@ func WithMetadata(meta string) ConnectOption {
 // out of date.
 //
 // For now, don't use this outside of official consumers.
-func WithChildClient(clientType, version string) ConnectOption {
+func WithChildClient(clientType, version string, comments ...string) ConnectOption {
 	return func(cfg *connectConfig) {
-		cfg.ClientInfo = append([]clientInfo{{clientType, version}}, cfg.ClientInfo...)
+		cfg.ClientInfo = append([]clientInfo{{clientType, version, comments}}, cfg.ClientInfo...)
```
jrobsonchase, line 195: Do we actually need multiple comment support? We can adjust the parser to accommodate server-side, but I'm not sure we really need more than just the one.

Willing to be convinced though!

natasha-jarus: this is just my cheaty way of getting an `Option<string>` type, but i figured while i was making bad api decisions i might as well give away multiple comments for free

jrobsonchase: Ahh, gotcha. Maybe instead of multiple comments, we use Mozilla's (or maybe everyone's? my browser's at any rate) convention and semicolon-space separate them in a single comment? Even if multiple comments are technically supported by both the RFC/backend, a single comment with multiple values feels more "expected"

jrobsonchase: e.g. `Mozilla/5.0 (X11; Linux x86_64; rv:109.0) Gecko/20100101 Firefox/112.0`

jrobsonchase (approved):

PR #97, opened 2023-05-26
Send opts.TLSTermination if terminateAtEdge is set

bobzilladev:


config/tls_termination.go
```diff
@@ -27,6 +27,7 @@ func (tt tlsTermination) ApplyTLS(cfg *tlsOptions) {
 		// TODO: implement this in the tunnel `Accept` call.
 		panic("automatic tls termination in-app is not yet supported")
 	case TLSAtEdge:
+		cfg.terminateAtEdge = true
 		cfg.KeyPEM = tt.key
 		cfg.CertPEM = tt.cert
```
jrobsonchase, line 32: One alternative would be to make sure that these are always non-nil here, but I think the extra boolean makes the intent clearer :+1:

config/tls.go
```diff
@@ -66,7 +68,9 @@ func (cfg *tlsOptions) toProtoConfig() *proto.TLSEndpoint {
 
 	opts.MutualTLSAtEdge = mutualTLSEndpointOption(cfg.MutualTLSCA).toProtoConfig()
 
-	if cfg.KeyPEM != nil && cfg.CertPEM != nil {
+	// When terminate-at-edge is set the TLSTermination must be sent even if the key and cert are nil,
+	// this will default to the ngrok edge's automatically provisioned keypair.
+	if cfg.terminateAtEdge || (cfg.KeyPEM != nil && cfg.CertPEM != nil) {
```
jrobsonchase, line 73: Let's reduce this to just `if cfg.terminateAtEdge { ... }`, since eventually we may support in-library termination where we definitely don't want to send the cert/key ngrok-side.

jrobsonchase (changes requested):

config/tls_termination.go
bobzilladev: It does look like I could set zero-length arrays and get the same server effect, if you'd rather I went that route.

config/tls.go
bobzilladev: Updated to just use the bool.

bobzilladev:
* main
  * **PR #97** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-go/97" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-go/97?utm_source=stack-comment).

jrobsonchase (approved):

PR #98, opened 2023-05-31
fix (and test) websockets

jrobsonchase:
Back in the day, Go didn't have very good cancellation semantics. I
mean, it still doesn't, but it's "better."

All of the old io interfaces lack a context argument to check
to see if their blocking call has been canceled. Therefore, if you
spawn off a "reader" task, how do you tell it to quit reading? You can
handle exiting the loop easy enough on your own, but what if the `Read`
method never returns? You still need a way to cancel _that_.

The solution? `Conn.SetDeadline` to "a long time ago!"

No, seriously:
```go
func (cr *connReader) abortPendingRead() {
	cr.lock()
	defer cr.unlock()
	if !cr.inRead {
		return
	}
	cr.aborted = true
	cr.conn.rwc.SetReadDeadline(aLongTimeAgo)
	for cr.inRead {
		cr.cond.Wait()
	}
	cr.conn.rwc.SetReadDeadline(time.Time{})
}
```

To be fair, its usage as a cancellation mechanism [is documented](https://pkg.go.dev/net#Conn), but it's still less-than-obvious.

Anyway, this `abortPendingRead` is in the callstack for the `Hijack`
method for the response writer, which is how websocket connections get
taken over for their nefarious non-http purposes. Because we didn't have
`SetDeadline` implemented for this buffer type, the in-flight reads were
never getting canceled, so this "abort" call never finished, and the
websocket handler was stuck waiting forever for a connection.

### DIFF
--- a/internal/muxado/buffer.go
+++ b/internal/muxado/buffer.go
@@ -131,13 +131,11 @@ func (b *inboundBuffer) SetDeadline(t time.Time) {
 	b.deadline = t
 	if timeout := time.Until(t); timeout > 0 {
 		b.startTimerLocked(timeout)
+	} else {
+		b.stopTimerLocked()
 	}
 
-	if t.IsZero() {
-		b.stopTimerLocked()
-	} else {
-		b.cond.Broadcast()
-	}
+	b.cond.Broadcast()
 
 	b.mu.Unlock()
 }
--- a/internal/muxado/buffer.go
+++ b/internal/muxado/buffer.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"io"
 	"io/ioutil"
+	"os"
 	"sync"
 	"time"
 )
@@ -25,8 +26,10 @@ type inboundBuffer struct {
 	cond sync.Cond
 	mu   sync.Mutex
 	bytes.Buffer
-	err     error
-	maxSize int
+	err      error
+	maxSize  int
+	deadline time.Time
+	timer    *time.Timer
 }
 
 func (b *inboundBuffer) Init(maxSize int) {
@@ -57,9 +60,47 @@ DONE:
 	return n, err
 }
 
+// Notify readers that the deadline has arrived.
+func (b *inboundBuffer) notifyDeadline() {
+	// It's important that the mutex is locked for this. It ensures that an
+	// in-flight timer can't broadcast before a reader gets to its condvar.Wait().
+	b.mu.Lock()
+	b.cond.Broadcast()
+	b.mu.Unlock()
+}
+
+// Start or reset the timer.
+// Must be called when the mutex is locked.
+func (b *inboundBuffer) startTimerLocked(timeout time.Duration) {
+	if b.timer == nil {
+		b.timer = time.AfterFunc(timeout, b.notifyDeadline)
+	} else {
+		b.timer.Reset(timeout)
+	}
+}
+
+// Stops a timer, if one is set.
+// Must be called while the mutex is locked.
+func (b *inboundBuffer) stopTimerLocked() {
+	if b.timer != nil {
+		b.timer.Stop()
+	}
+}
+
 func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 	b.mu.Lock()
 	for {
+		// If the deadline is set, we need to take it into account
+		if !b.deadline.IsZero() {
+			// If the deadline is in the past, bail out.
+			// SetDeadline will ensure that we get woken back up if it expires.
+			if time.Until(b.deadline) < 0 {
+				n = 0
+				err = os.ErrDeadlineExceeded
+				break
+			}
+		}
+
 		if b.Len() != 0 {
 			n, err = b.Buffer.Read(p)
 			break
@@ -68,6 +109,7 @@ func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 			err = b.err
 			break
 		}
+
 		b.cond.Wait()
 	}
 	b.mu.Unlock()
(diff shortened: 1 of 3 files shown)


jrobsonchase:
Failing test can be seen by checking out the pre-fix commit and running:
```
NGROK_TEST_ALL=1 go test --run=Websockets --timeout=5s
```

internal/muxado/buffer.go
```diff
@@ -68,7 +98,24 @@ func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 			err = b.err
 			break
 		}
+
+		// If the deadline is set, we need to take it into account
+		if !b.deadline.IsZero() {
```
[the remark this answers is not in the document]
jrobsonchase: That's true, but since `Read` is pretty much always called in a loop, it'll just get called again immediately after and return the error. 

```go
n, err = conn.Read(buf) // non-zero, ErrDeadlineExceeded

// vs

n, err = conn.Read(buf) // non-zero, nil
n, err = conn.Read(buf) // 0, ErrDeadlineExceeded
```

Feels like they're more or less the same thing, but with less opportunity for bad consumers to forget to handle the non-zero/non-nil-error case.

I guess there's the possibility of the buffer getting filled faster than the reader can drain it, which would cause the deadline to never get seen?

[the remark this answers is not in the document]
jrobsonchase: OK, moved the deadline check to the start of the loop

jrobsonchase:
Can't figure out why this stupid test fails in CI but works fine for me locally :upside_down_face:

internal/muxado/buffer.go
```diff
@@ -68,7 +113,12 @@ func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 			err = b.err
 			break
 		}
+
+		b.startTimerLocked(timeout)
```
nikolay-ngrok, line 117: isn't this going to start the timer in all cases (even when we don't have a deadline), but also use `0` as a timeout, effectively waking up the goroutine all the time? And then continue to set/wake at a rapid pace?

jrobsonchase: Ah! I think this is the bug that's causing my test failures. I was also shadowing rather than setting the `timeout` variable. Thanks Go.

internal/muxado/buffer.go
```diff
@@ -82,25 +125,21 @@ func (b *inboundBuffer) SetError(err error) {
...
-		delay := t.Sub(time.Now())
+	b.mu.Lock()
 
-		if b.timer != nil {
-			b.timer.Stop()
-		}
+	// Set the deadline and notify any readers that they need to take heed.
+	// They'll figure out all of the timer management for us.
+	b.deadline = t
+	b.startTimerLocked(time.Until(t))
```
nikolay-ngrok, line 133: do we check if this is positive to avoid setting up an expired timer and `Broadcast`ing twice?

jrobsonchase: Shouldn't matter - readers should be resilient to spurious wakeups, but not a bad thought :+1:

nikolay-ngrok: Also saves creating the `Timer` for `SetDeadline(time.Zero)` 🤷

internal/muxado/buffer.go
```diff
@@ -77,30 +119,34 @@ func (b *inboundBuffer) Read(p []byte) (n int, err error) {
...
+	b.mu.Lock()
 
-		// how long until the deadline
-		delay := t.Sub(time.Now())
+	// Set the deadline and notify any readers that they need to take heed.
+	// They'll figure out all of the timer management for us.
+	b.deadline = t
+	if timeout := time.Until(t); timeout > 0 {
+		b.startTimerLocked(timeout)
+	}
```
nikolay-ngrok, line 134: maybe a bit nicer would be to stop the timer as an else here 🤔

nikolay-ngrok: and then `Broadcast` all the time 🤷

internal/muxado/buffer.go
```diff
@@ -57,9 +60,47 @@ DONE:
...
+
 func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 	b.mu.Lock()
 	for {
+		// If the deadline is set, we need to take it into account
+		if !b.deadline.IsZero() {
+			// If the deadline is in the past, bail out.
+			// SetDeadline will ensure that we get woken back up if it expires.
+			if time.Until(b.deadline) < 0 {
+				n = 0
```
nikolay-ngrok, line 98: technically, no need for `n = 0`, right?

nikolay-ngrok (approved):
I think this works 👍

internal/muxado/buffer.go
jrobsonchase: I never fully trust these named return values, so I like being explicit just on the off-chance that someone later breaks these assumptions today :stuck

internal/muxado/buffer.go
```diff
@@ -68,6 +109,7 @@ func (b *inboundBuffer) Read(p []byte) (n int, err error) {
 			err = b.err
 			break
 		}
+
 		b.cond.Wait()
 	}
 	b.mu.Unlock()
```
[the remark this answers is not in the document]
jrobsonchase: Mmm, maybe. But I'm a little hesitant to change up the control flow just because it's another thing that I'll have to re-reason about :grin:

internal/muxado/buffer.go
jrobsonchase: I think we end up doing an unneeded broadcast on deadline == zero, but that's probably ok :shrug:

PR #99, opened 2023-06-01
implement buffer deadline to support websockets

nikolay-ngrok:


nikolay-ngrok:
This is just an experiment using channel locks, closing in favor of #98

PR #103, opened 2023-06-22
WithStopHandler CloseWrite

abakum:
Without CloseWrite handleConn never return (in my app)

### DIFF
--- a/examples/ngrok-lite/main.go
+++ b/examples/ngrok-lite/main.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net"
 	"os"
+	"time"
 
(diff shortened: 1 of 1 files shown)


bobzilladev:
Hello! The example being modified by this pull request isn't around anymore, but there is now [an example](https://github.com/ngrok/ngrok-go/blob/main/examples/http-full/main.go) with all HTTP and Session options including `WithStopHandler`, so folks can see an example of how that works, albeit without the connection closing.

Thank you for providing this illustration to help anyone who started with this example and wanted to implement the stop functionality! Going to close this issue now due to the above, thank you for the feedback.

PR #112, opened 2023-08-17
Expose ngrok error codes in error types

### LINKED ISSUE #111: Implement `error_code` for RPC errors
Similar to the implementation for [ngrok-rust](https://github.com/ngrok-private/ngrok/issues/21498), we'll need to augment the error structs within ngrok-go to also parse and return the associated error code.

Wonderful-Space:
Resolves #111

### DIFF
--- a/errors.go
+++ b/errors.go
@@ -3,7 +3,13 @@ package ngrok
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 )
+
+type ngrokError interface {
+	Error() string
+	ErrorCode() string
+}
 
 // Errors arising from authentication failure.
 type errAuthFailed struct {
@@ -34,6 +40,19 @@ func (e errAuthFailed) Is(target error) bool {
 	return ok
 }
 
+func (e errAuthFailed) ErrorCode() string {
+	errMsg := e.Inner.Error()
+
+	// Find the number in the format of ERR_NGROK_(\d+).
+	reg := regexp.MustCompile(`ERR_NGROK_(\d+)`)
+	matches := reg.FindStringSubmatch(errMsg)
+	if len(matches) == 2 {
+		return matches[1]
+	}
+
+	return ""
+}
+
 // The error returned by [Tunnel]'s [net.Listener.Accept] method.
 type errAcceptFailed struct {
 	// The underlying error.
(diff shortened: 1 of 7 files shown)


errors.go
```diff
@@ -3,8 +3,14 @@ package ngrok
 import (
 	"fmt"
 	"net/url"
+	"regexp"
 )
 
+type ngrokError interface {
```
euank, line 9: Since this is unexported and unused in the codebase, this actually doesn't have any benefit, right?

errors.go
```diff
@@ -34,6 +40,19 @@ func (e errAuthFailed) Is(target error) bool {
 	return ok
 }
 
+func (e errAuthFailed) ErrorCode() string {
+	errMsg := e.Inner.Error()
+
+	// Find the number in the format of ERR_NGROK_(\d+).
+	reg := regexp.MustCompile(`ERR_NGROK_(\d+)`)
```
euank, line 47: We shouldn't compile the same regexp over and over.

It also seems wrong that we don't have a proper typed way to get just the errcode part here.
I guess the server isn't providing a properly structured error code?

Shouldn't we be changing the server / lib/tunnel implementation to give us the errcode in a properly structured way, and then implementing the agent sdk changes based on that?

errors.go
```diff
@@ -34,6 +40,19 @@ func (e errAuthFailed) Is(target error) bool {
...
+	matches := reg.FindStringSubmatch(errMsg)
+	if len(matches) == 2 {
+		return matches[1]
```
euank, line 50: It feels surprising that this returns, say, "123" instead of "ERR_NGROK_123".

The reason for that is consistency with the go api library: https://pkg.go.dev/github.com/ngrok/ngrok-api-go/v5#Error

There, we have `Error.ErrorCode`, which is `ERR_NGROK_123`, and then we have the helper method `ngrok.IsErrorCode(err, 400)`.

It feels like we should have consistency in how users think about errors, or at least have an explanation for why we're diverging if we do.

inconshreveable: > There, we have Error.ErrorCode, which is ERR_NGROK_123, and then we have the helper method ngrok.IsErrorCode(err, 400).

for some additional context as the author of ngrok-go-api: if i were to rev the ngrok api go client library, i'd change that function to be `ngrok.IsErrorCode(error, ...string)`. this further supports your suggestion. the reason being that 
we wanted to one day stop namespacing errors with numbers (all api errors are `ERR_NGROK_4XXX`) and instead move to namespacing with human-readable strings, e.g. `ERROR_NGROK_API_12` 

thanks for picking this up carl, this improvement is much appreciated!

errors.go
jrobsonchase: Ideally, yes, but I'd prefer to avoid making piecemeal adjustments to the current RPC request/response objects that we have to continue supporting for various agent versions. Someday (hopefully soon?) I want to overhaul them into protobuf messages instead of these hand-rolled json objects, and _then_ we'll make sure to send structured errors.

errors.go
jrobsonchase: Oh, this might have been my bad. Carl is just porting my work in ngrok/ngrok-rust over to go, and I omitted the `ERR_NGROK_` prefix from the error code, since it seemed implied. I'm cool with adjusting it to have the prefix if we want to support namespacing in the future :+1:

errors.go
Wonderful-Space: Yeah, definitely my bad. Will make it exported.

jrobsonchase (approved):

PR #113, opened 2023-08-21
push error parsing down to the proto package

jrobsonchase:


session.go
```diff
@@ -604,7 +604,6 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
 		if err != nil {
 			remote := false
 			if resp.Error != "" {
-				err = errors.New(resp.Error)
```
jrobsonchase, line 607: This was an oopsie on my part - the internal Session is already setting the `err` for us.

Wonderful-Space (approved):

PR #120, opened 2023-09-14
Update forwarding API

Wonderful-Space:
Resolves https://github.com/ngrok-private/ngrok/issues/21878

**HTTP lite example**
image
image

**Forward lite example**
image
image
image


session.go
```diff
@@ -805,6 +803,43 @@ func (s *sessionImpl) Warnings() []error {
 		return []error{(*AgentVersionDeprecated)(deprecated)}
 	}
 	return nil
+
+}
+
+func (s *sessionImpl) ListenAndForward(ctx context.Context, url *url.URL, cfg config.Tunnel) (Forwarder, error) {
+	tun, err := s.Listen(ctx, cfg)
```
jrobsonchase, line 810: We should be setting the forwards_to in the config before calling listen

session.go
```diff
@@ -51,6 +51,15 @@ type Session interface {
...
 	Warnings() []error
 
+	// ListenAndForward creates a new Tunnel which will listen for new inbound
+	// connections. Connections on this tunnel are automatically forwarded to
+	// the provided URL.
+	ListenAndForward(ctx context.Context, backend *url.URL, cfg config.Tunnel) (Forwarder, error)
+
+	// ListenAndServeHTTP creates a new Tunnel to serve as a backend for an HTTP server. Connections will be
+	// forwarded to the provided HTTP server.
+	ListenAndServeHTTP(ctx context.Context, cfg config.Tunnel, server *http.Server) (Tunnel, error)
```
jrobsonchase, line 61: This should probably return a Forwarder as well, so as not to expose the Tunnel's Accept call and cause weirdness with the HTTP server.

jrobsonchase, line 61: Should we maybe also have a `ListenAndHandleHTTP` that takes an `http.Handler`, and wraps it in a default http.Server for you? Kinda like with the now-deprecated options.

session.go
```diff
@@ -779,24 +789,12 @@ func (s *sessionImpl) Listen(ctx context.Context, cfg config.Tunnel) (Tunnel, er
...
-		HTTPServer() *http.Server
-	}); ok {
-		if srv := httpServerCfg.HTTPServer(); srv != nil {
-			go func() {
-				_ = srv.Serve(t)
-			}()
-		}
-	}
-
-	return t, nil
```
jrobsonchase, line 799: Hmm... not sure how I feel about actively *breaking* this behavior, even if the methods are deprecated. What if we went ahead and created our waitgroup here and added it to the `tunnelImpl`, and then if `ListenAndServe` is called afterward and we find that a server has already been started in `Listen`, we effectively no-op and use the existing server? And probably spit out a warning.

jrobsonchase (dismissed):
Looking pretty good! A few minor notes, and we'll need to support proxy proto/header rewriting at some point, but it's as good of a start as we've got in Rust :+1:

session.go
Wonderful-Space: Yeah, I had that thought, wasn't sure if we wanted to supported `http.Handler` or not. I'll add it in 👍🏼

session.go
Wonderful-Space: Ah I see, if they call `Listen` directly. 👍🏼 Can do

PR #131, opened 2023-10-19
Adds TLS Renegotiation to the backend

bobzilladev:
Tested against a server requiring renegotiation. Using `Once` since this is validated.


nikolay-ngrok (approved):
👍

PR #136, opened 2023-11-03
Handle StopTunnel request

bobzilladev:
Handle a new `StopTunnel` message, which will close the tunnel, causing an error on the tunnel's `Accept()`, and maintain the running session.


internal/tunnel/client/session.go
```diff
@@ -67,6 +67,10 @@ type Session interface {
 	// Latency updates
 	Latency() <-chan time.Duration
 
+	// Close the tunnel with this clientID, with an error that will be reported
+	// from the tunnel's Accept() method.
+	CloseTunnel(clientID string, err error) error
```
jrobsonchase, line 72: Let's not make this part of the public interface. I think I saw some SDK feedback that we should remove the other similiar ones from rust/node/python.

session.go
```diff
@@ -99,6 +103,9 @@ type SessionHeartbeatHandler func(ctx context.Context, sess Session, latency tim
 // ServerCommandHandler is the callback type for [WithStopHandler]
 type ServerCommandHandler func(ctx context.Context, sess Session) error
 
+// StopTunnelHandler is the callback type for [WithStopTunnelHandler]
+type StopTunnelHandler func(ctx context.Context, sess Session, clientID string, err error) error
```
jrobsonchase, line 107: Do we need a user-assignable handler for this? What do you get out of this that you wouldn't be able to discern from the `Tunnel.Accept` method returning an error?

If we _were_ to allow users to set this, it feels a little odd that we still close the tunnel outside of the user-provided handler. It differs from the other handlers that we offer (Stop, Restart, Update), which are no-ops if the user doesn't set them. Seems unexpected that this one doesn't actually let you change the behavior.

examples/ngrok-forward-lite/main.go
```diff
@@ -57,19 +57,32 @@ func main() {
...
-	return fwd.Wait()
+		l.Log(ctx, ngrok_log.LogLevelInfo, "tunnel created", map[string]any{
+			"url": fwd.URL(),
+		})
+
+		err = fwd.Wait()
+		if err == nil {
+			return nil
+		}
+		l.Log(ctx, ngrok_log.LogLevelWarn, "tunnel accept error. now setting up a new listener.",
```
jrobsonchase, line 85: > ... tunnel ...

:wink:

internal/tunnel/client/reconnecting.go
```diff
@@ -167,7 +167,9 @@ func (s *reconnectingSession) receive() {
 		// accept the next proxy connection
 		proxy, err := s.raw.Accept()
 		if err == nil {
-			go s.handleProxy(proxy)
+			if proxy != nil {
+				go s.handleProxy(proxy)
+			}
```
jrobsonchase, line 172: Is this ever actually the case? Seems like we should never see a `nil, nil` return from `Accept`. One or the other should probably always be set.

jrobsonchase (commented):
Overall approach looks good, but I think we should remove the public-facing `CloseTunnel` method as well as the user-definable `StopTunnel` handler.

Still need to go through the internal agent change to see how we're actually using it. Might have some suggestions there.

session.go
```diff
@@ -959,3 +984,16 @@ func (rc remoteCallbackHandler) OnUpdate(_ *proto.Update, respond tunnel_client.
...
 }
+
+func (rc remoteCallbackHandler) OnStopTunnel(stopTunnel *proto.StopTunnel, respond tunnel_client.HandlerRespFunc) error {
+	err := &ngrokError{Message: stopTunnel.Message, ErrCode: stopTunnel.ErrorCode}
+	if rc.stopTunnelHandler != nil {
+		// allow user to return an error and instigate a reconnect
+		user_err := rc.stopTunnelHandler(context.TODO(), rc.sess, stopTunnel.ClientID, err)
+		if user_err != nil {
+			return user_err
+		}
```
jrobsonchase, line 995: OK, having looked at the agent side of this, I'm not sure this is really what we want.

As I understand it, this would initiate a "hot" reconnect where the session sends its cookie in the auth request and tries to rebind all of its tunnels with the hostports that it was given the last time they were bound. Since we expect this to only really occur (right now at any rate) when this agent lost a race to get an ephemeral port assignment, that reconnect would just loop on trying to rebind that tunnel that it already lost. This would just throw the agent agent into a reconnect loop until someone comes along and ctrl+c's it.

We can either retain the original configuration for the tunnel, try to detect this condition, and attempt a new bind with the same config, or we can skip trying to handle reconnecting the session here altogether and leave it up to the caller to decide what to do with their now-dead tunnel. The former sounds like a whole lot of not-very-fun plumbing to me, plus it would mean that the existing Tunnel object would get a new URL, which feels wrong. It's kinda signaled by this handler if we decided to keep it, but then the user has to maintain the ID mappings for all of their bound tunnels if they want to plumb the "this tunnel now has a new URL" context to wherever. 

So I still think we skip this handler altogether along with any attempt at reconnecting the session here, and let the `Accept` error be the way this gets signaled. Then on the agent side, we can use that error to either trigger a full session restart to get brand new tunnel URLs, or it can attempt to bind a new tunnel with the same config as the dead one.

bobzilladev: tl;dr: Sure, I'll start switching the agent to a cold reconnect, or rebind with same tunnel-config, depending on which is less complex. That will eliminate the need for `ngrok-go`'s `stopTunnelHandler`.

Aye, you have it right, the agent would go into a "hot" reconnect loop, just like the current closing of the session, under the condition of duplicate tcp address.

For context on why it's implemented this way currently: The goal of the [original issue](https://github.com/ngrok-private/ngrok/issues/22761) was to improve the UX of session close by getting an informative message to the user (and why the original plan was just a status message but was extended to the more specific StopTunnel). Coming up with a plan for changing out hostports on a tunnel-config and broadcasting that out was for a [separate future issue](https://github.com/ngrok-private/ngrok/issues/22597). So agent behavior of hot reconnect was maintained, but with an error message for the user.

But we can make other decisions here too. In all cases we can probably surface an error on the UI that will stick, so the menu of options seem to be:

1. Hot reconnect. If for some reason the other tcp tunnel goes away this will self-heal. Currently implemented.
2. Cold reconnect
    a. If it's an ephemeral tcp address a new one will be provided, which could also update in the UI, but there's no callback to get information that this happened into other software.
    b. If it's a user-specified tcp address this will go into a reconnect loop like option 1, unless the other conflicting tunnel goes away.
3. Tunnel rebind with same config. Ends up basically same as cold reconnect for this case, but might be more complicated to implement.
4. No reconnect, so the agent acts like the SDK's: the session goes on, as do any other tunnels (which are unlikely to exist in agent case). No possibility of self-healing, but the likelihood of that is small anyway.

1 is implemented, you're leaning towards 2 or 3. Between 2 and 3 I'd argue for 2 (cold reconnect) since most agents only have the one tunnel, and that's likely less complex (but we'll see).

Between 1 (hot) and 2 (cold) depends on if it's ok to swap out a hostport under someone. I didn't think that was an option at this time, but it sounds like it could be. I'm not passionate either way, this is a rare condition, and we'll make an error message surface regardless, so I'll start pursuing 2/3.

CC'ing product in case there are stronger concerns there: @nijikokun, @salilsub

internal/tunnel/client/reconnecting.go
bobzilladev: The `nil, nil` is for the "don't start a proxy and don't reconnect" case, which is the default for handling the `StopTunnel` message. Will just hit the `continue` below and go back to blocking on accepting more from muxado. Alternative could be an "everything is ok" error, but that's a little odd too.

internal/tunnel/client/session.go
bobzilladev: Like you mention below, left this `CloseTunnel` since it's on the internal client package.

session.go
```diff
@@ -66,6 +66,10 @@ type Session interface {
 	// forwarded to a new HTTP server and handled by the provided HTTP handler.
 	ListenAndHandleHTTP(ctx context.Context, cfg config.Tunnel, handler *http.Handler) (Forwarder, error)
 
+	// Close the tunnel with this clientID, with an error that will be reported
+	// from the tunnel's Accept() method.
+	CloseTunnel(clientID string, err error) error
+
```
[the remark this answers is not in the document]
bobzilladev: Removed!

session.go
bobzilladev: Removed! User can't force a hot reconnect now, which was the main reason for it to exist.

examples/ngrok-forward-lite/main.go
bobzilladev: Removed!

internal/tunnel/client/reconnecting.go
bobzilladev: Removed!

session.go
bobzilladev: This is removed, the agent will handle this in a different way, as suggested.

PR #137, opened 2023-11-05
when the connection is broken, join() without Close() freezes for ten…

abakum:
…s of seconds

### DIFF
--- a/forward.go
+++ b/forward.go
@@ -53,15 +53,17 @@ func (fwd *forwarder) Wait() error {
 // compile-time check that we're implementing the proper interface
 var _ Forwarder = (*forwarder)(nil)
 
-func join(ctx context.Context, left, right io.ReadWriter) {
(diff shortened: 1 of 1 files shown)


abakum:
```
func join(ctx context.Context, left, right net.Conn) {
	g, _ := errgroup.WithContext(ctx) // when ctx is canceled (on WithStopHandler or WithDisconnectHandler ) interrupts both io.Copy 
	g.Go(func() error {
		_, err := io.Copy(left, right)
		left.Close() // on left disconnection interrupts io.Copy(right, left) 
		return err
	})
	g.Go(func() error {
		_, err := io.Copy(right, left)
		right.Close() // on right disconnection interrupts io.Copy(left, right) 
		return err
	})
	g.Wait()

}
```

euank:
I carried forward and merged an equivalent change over in #163, closing this one as resolved via that PR.
Thanks!

PR #141, opened 2023-11-21
ngrok: add WithAppProtocol

### LINKED ISSUE #139: Support `WithAppProtocol` for HTTP and Labeled listeners
This is to support http/2, and potentially other change-of-protocols that may occur.

See ngrok-private/ngrok#23863

jrobsonchase:
Resolves #139

Adds the `ForwardsProto` field to the `Bind` request, as well as `WithAppProtocol` as an HTTP/labeled option to set it when creating a tunnel.


jrobsonchase:
* main
  * **PR #141** <a href="https://app.graphite.dev/github/pr/ngrok/ngrok-go/141?utm_source=stack-comment-icon" target="_blank">Graphite</a>  👈

This [stack of pull requests](https://stacking.dev/?utm_source=stack-comment) is managed by [Graphite](https://app.graphite.dev/github/pr/ngrok/ngrok-go/141?utm_source=stack-comment).

jrobsonchase:
Added the tentative user-facing config options, still need agreement on them though as well as documentation of valid values.

config/app_protocol.go
```diff
@@ -0,0 +1,24 @@
...
+func (ap appProtocol) ApplyLabeled(cfg *labeledOptions) {
+	cfg.commonOpts.ForwardsProto = string(ap)
+}
+
+// WithAppProtocol declares the protocol that the upstream service speaks.
+// This may be used by the ngrok edge to make decisions regarding protocol
+// upgrades or downgrades.
+//
+// Currently, `http2` is the only valid string, and will cause connections
+// received from HTTP endpoints to always use HTTP/2.
```
nikolay-ngrok, line 18: is this contradicting:
> // the L7 protocol the upstream service is expecting (one of '', 'h1', 'h2c')

nikolay-ngrok (commented):
looks straightforward 👍

config/app_protocol.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (ap appProtocol) ApplyLabeled(cfg *labeledOptions) {
+	cfg.commonOpts.ForwardsProto = string(ap)
+}
+
+// WithAppProtocol declares the protocol that the upstream service speaks.
+// This may be used by the ngrok edge to make decisions regarding protocol
+// upgrades or downgrades.
+//
+// Currently, `http2` is the only valid string, and will cause connections
```
bobzilladev, line 17: This mentions `http2` as the only legal value, but the proto msg has `ForwardsProto string    // the L7 protocol the upstream service is expecting (one of '', 'h1', 'h2c')`. I don't think there's any translation or validation going on in here, so just wanted to double-check if either of these comments might be incorrect.

bobzilladev: heh, what niki said.

salilsub:
@jrobsonchase What does the error-logic look like if someone puts an invalid option? For example, if someone used `--app-protocol moose`, what would they see?

config/app_protocol.go
jrobsonchase: Yeah, this is kinda a product question. the `'h1', 'h2c'` bits were copied from the internal change, and it's what we're expecting on the backend, but the product spec uses `http1, http2`.

I can update the comment on the field to match though.

jrobsonchase:
The configuration should fail validation and our services should return some sort of "invalid config" error. We _could_ validate locally before sending it off, but I wouldn't. That way, we don't have to keep things synchronized between the sdk and our backend if new protocols are added. Kinda the same deal as with region validation (or lack thereof :stuck_out_tongue:)

bobzilladev (approved):
+1'd now that it's consistent `http1`/`http2`

salilsub (approved):

PR #146, opened 2023-12-05
config: add support for policies on http/tcp/tls tunnels

rkolavo:
The policies module allows users to attribute a set of rules that will be conditionally applied to the traffic associated with an edge. A single policy may have 0 to many cel expressions that evaluate to a boolean, and 1 to many actions that will execute if the aforementioned boolean is true.

Policies are classified as either inbound (apply to connections before they reach a backend) or outbound (apply to connections after they've come back from a backend).


config/policies.go
```diff
@@ -0,0 +1,152 @@
...
+	opts []policiesOption
+}
+type policyBuilder struct {
+	opts []policyOption
+}
+type actionBuilder struct {
+	opts []actionOption
+}
+
+// Add the provided policies to the ngrok edge
```
Megalonia, line 35: "New Policies adds the ... "

jrobsonchase:
We try to stick with the [functional options](https://golang.cafe/blog/golang-functional-options-pattern.html) pattern where possible, but the API for policies looks to be using builders instead

rkolavo:
> We try to stick with the [functional options](https://golang.cafe/blog/golang-functional-options-pattern.html) pattern where possible, but the API for policies looks to be using builders instead. See the [OAuth](https://github.com/ngrok/ngrok-go/blob/main/config/oauth.go) configuration as an example.

womp, my bad, I had a bit of changes from trying both patterns out. Think I have the fully functional pattern built out now. Thanks @jrobsonchase

config/policies.go
```diff
@@ -0,0 +1,156 @@
...
+// The order in which policies are added is respected at runtime.
+func WithOutboundPolicy(opts ...PolicyOption) PoliciesOption {
+	return func(p *policies) {
+		outP := &policy{}
+		for _, opt := range opts {
+			opt(outP)
+		}
+		p.Outbound = append(p.Outbound, outP)
+	}
+}
```
jrobsonchase, line 63: Let's drop the `WithPolicies` option and make `WithInboundPolicy` and `WithOutboundPolicy` the top-level endpoint options.

config/policies.go
```diff
@@ -0,0 +1,156 @@
...
+
+// WithName sets the provided name on this policy
+func WithName(name string) PolicyOption {
```
jrobsonchase, line 66: Within a generic `config` package, options like this are themselves far too generic. Let's insert `Policy` between the `With` and *thing* for all of the `PolicyOption`s, e.g. `WithPolicyName`

config/policies.go
```diff
@@ -0,0 +1,156 @@
...
+		act := &action{}
+		for _, opt := range opts {
+			opt(act)
+		}
+		p.Actions = append(p.Actions, act)
+	}
+}
+
+// WithType sets the provided type for this action. Type must be specified.
+func WithType(typ string) ActionOption {
```
jrobsonchase, line 92: Ditto the above comment about `PolicyOption`s. Didn't even notice at first glance that these were *Action* options lol

jrobsonchase (commented):
*puts on Product hat* :tophat:

config/policy.go
```diff
@@ -0,0 +1,254 @@
...
+	return a
+}
+
+// WithActionType sets the provided type for this action. Type must be specified.
+func WithPolicyActionType(t string) policyActionOption {
+	return optionFunc[*action](func(a *action) { a.Type = t })
+}
+
+// WithConfig sets the provided map as the configuration for this action
+func WithPolicyActionConfig(c map[string]any) policyActionOption {
```
rkolavo, line 157: Our SDK design says to accept a json string here, but is that really what we want? Or should we accept a map like this?

config/policy.go
```diff
@@ -0,0 +1,254 @@
...
+
+// supports action as an option of a policy rule
+func (a action) apply(p *policyRule) {
+	p.Actions = append(p.Actions, a)
+}
+
+// an option that is applicable to the specified type
+type option[T any] interface {
+	apply(T)
+}
```
rkolavo, line 192: Using this interface allowed me to have policyRule and action be both structs that can be converted to json _and_ be 'options' of policy and policyRule respectively. I can't tell if I got too "clever" with it to the point that it's not easily follow-able.

config/policy.go
```diff
@@ -0,0 +1,248 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("failed to convert to json with error: %s", err.Error()))
+	}
+
+	return string(bytes)
+}
+
+// WithPolicyConfig configures this edge with the provided policy configuration
+// passed as a json string and overwrites any prevously-set traffic policy
+// https://ngrok.com/docs/http/traffic-policy
```
bobzilladev, line 62: This mentions `traffic-policy`, wanted to double check that that name is current (I just haven't heard `traffic` before here, and that docs link isn't live).

bobzilladev (approved):
One question about "traffic policy" naming/docs, but change lgtm.

config/policy.go
rkolavo: That is indeed the latest on what we are calling the module itself "Traffic Policy" but all the configuration structures just refer to it as "policy"

PR #148, opened 2023-12-23
feat(log): change slog import path

dwisiswant0:


jrobsonchase (approved):

jrobsonchase:
Actually, I think we need to update the go version in the go.mod before merging this, since `log/slog` was added in `1.21`.

PR #153, opened 2024-01-22
Update the slog import path for go 1.21

jrobsonchase:
Supersedes #148 

Pulls in @dwisiswant0's change and updates the go version to match.


.gitignore
```diff
@@ -1,3 +1,4 @@
+/go.work.sum
```
jrobsonchase, line 1: We already have go mod sums - this one is just noise.

OfTheDelmer (approved):

PR #156, opened 2024-02-05
Traffic Policy revisions

rkolavo:
- Breaks out the traffic policy structs into their own package and exports them. 
- Changes the top-level endpoint options to be `WithPolicyString()` (which accepts a json string) and `WithPolicy` (which accepts a `trafficpolicy.Policy` instance
- Removed nested functional options from within the Policy option.


trafficpolicy/policy.go
```diff
@@ -0,0 +1,91 @@
...
+	return marshalJSON(p)
+}
+
+func (p Rule) MarshalJSON() (string, error) {
+	return marshalJSON(p)
+}
+
+func (p Action) MarshalJSON() (string, error) {
+	return marshalJSON(p)
+}
```
rkolavo, line 57: I actually don't think naming these `MarshalJSON()` here is the right move after all since that's the function name used to override a type's marshaling behavior (see what I mean above for ActionConfigString). Leaning towards keeping these as `ToJSON()` or even just `JSON()`

trafficpolicy/policy.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (p Rule) MarshalYAML() (string, error) {
+	return marshalJSON(p)
+}
+
+func (p Action) MarshalYAML() (string, error) {
+	return marshalJSON(p)
+}
```
rkolavo, line 81: We wanted to support having the nested rule or action object convertible to json, but is this applicable/useful for yaml as well?

trafficpolicy/policy.go
```diff
@@ -0,0 +1,91 @@
...
+
+	if err := enc.Encode(o); err != nil {
+		return "", err
+	}
+
+	return b.String(), nil
+}
+
+func (p Policy) MarshalYAML() (string, error) {
+	return marshalJSON(p)
```
OfTheDelmer, line 72: should this utilize the `marshalYAML`?

rkolavo: YES

trafficpolicy/policy.go
nijikokun: Does that change consistency with the other parts of the API? Are there some places where it's called `MarshalJSON` and others where it is `ToJSON`? That would be my only concern. If not, then I'm fine calling it `ToJSON` or `JSON`.

trafficpolicy/policy.go
```diff
@@ -0,0 +1,91 @@
...
+	raw := json.RawMessage([]byte(s))
+
+	return json.Marshal(raw)
+}
+
+func (cfg *ActionConfigString) UnmarshalJSON(data []byte) error {
+	s := ActionConfigString(data)
+	*cfg = s
+	return nil
+}
```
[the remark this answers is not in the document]
nijikokun: By "this type" do you mean `ActionConfigString`?

trafficpolicy/policy.go
nijikokun: I could see myself saving them as separate yaml documents like mentioned earlier and then importing them later into various locations. Especially after we get the sugar methods for actions like `.Compression()`

trafficpolicy/policy.go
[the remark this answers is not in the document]
rkolavo: yes

trafficpolicy/policy.go
rkolavo: It would be consistent to the user of this package to always use `ToJSON` or `JSON` — the custom type `ActionConfigString` would never have its `MarshalJSON()` or `MarshalYAML()` function called except by the json/yaml encoders

rkolavo: We landed on `JSON()` and `YAML()`

CHANGELOG.md
```diff
@@ -1,3 +1,13 @@
+## 1.8.1
+Enhancements: 
+
+- Provides access to structs for building a Traffic Policy configuration
+
+Breaking changes:
+
+- Renames pre-release option `WithPolicyConfig` to `WithPolicyString`
+- Changes signature of `WithPolicy` option to accept the newly exposed structs for building a Traffic Policy configuration
```
jrobsonchase, line 9: There shouldn't be breaking changes in a non-major version release, much less a patch version. We can get away with it if we have deprecated aliases for new names. The signature change is trickier, but we can make the new `Policy` struct implement `PolicyOption`.

We'll obviously also need to leave the `WithPolicy*` functions in place, but can add deprecations to them.

CK-Ward: What are you suggesting? a 4.0.0 major version release? just to be clear

rkolavo: Agh yes, sorry meant to circle back on the versioning here. I did not think this was "right" but had mixed feelings since this was not a publicly available feature yet. I think I am in favor of making the leap to v2 here.

jrobsonchase: I'd really prefer _not_ to bump to v2. It's something that has to be manually handled by users, changes import paths, etc. 

And we should think _very carefully_ about what breaking changes we want to introduce then, and try to do them all in one go rather than make a major bump for each one.

nijikokun: Is anyone using this, it's not publicly available yet.

jrobsonchase: [Yes it is.](https://pkg.go.dev/golang.ngrok.com/ngrok@v1.8.0/config#WithPolicy)

nijikokun: It's currently behind a feature flag and undocumented.

jrobsonchase: OK, I guess it's almost OK to break these then. But this could still occur:

* Someone updated to 1.8.0 between when it was cut an now
* We announce Policies and turn the feature flag on
* They think "oh cool, lemme try that out," and use the version they've already got in their go.mod.
* Time passes, and they run go module updates. Suddenly their previously-working code fails to compile.

rkolavo: based on the conversation in slack, it sounds like we are making a very rare (never to be seen again!) exception here... are we okay with the current versioning? Any changes requested to verbiage in the changelog here?

jrobsonchase: Nope, fine as-is

jrobsonchase (approved):

PR #158, opened 2024-02-21
multiple session legs support

bobzilladev:
Adding support for multiple session legs.

This adds two configuration options:

- `WithAdditionalServers(addrs []string)`
- `WithMultiLeg(enable bool)`

If multi-leg is enabled, then the `ConnectAddresses` in the Auth response of the first session are used to determine what additional sessions should be set up. The `ReconnectingSession` will accumulate more sessions on connect, then return as normal. `ReconnectionSession` will handle fan-out of calls to the underlying `tunnel_client.Session` as before, except now there are multiple of them.

If the first connection is made using least-latency, it will lock in to that `ConnectAddress` so reconnects will go back to the same region. Heartbeat callbacks are done for each underlying session, including appropriate region and latency information.

### DIFF
--- a/internal/tunnel/client/reconnecting.go
+++ b/internal/tunnel/client/reconnecting.go
@@ -382,7 +382,9 @@ func (s *reconnectingSession) connect(acceptErr error, connSession *session) err
 		// check if more sessions need to be established
 		sendStateChange := true
 		if desiredLegs > len(s.sessions) {
-			// set up the next connection
+			// set up the next connection. additional sessions will
+			// continue to chain on from there until all legs are
+			// established
 			s.createTunnelClientSession(s)
 			// not done with initial setup yet
 			sendStateChange = false
--- a/internal/tunnel/client/reconnecting.go
+++ b/internal/tunnel/client/reconnecting.go
@@ -99,14 +99,14 @@ type reconnectingSession struct {
 	stateChanges chan<- error
 	clientID     string
 	cb           ReconnectCallback
-	swapper      *swapRaw
-	*session
+	sessions     []*session
+	log.Logger
 }
 
-type RawSessionDialer func() (RawSession, error)
-type ReconnectCallback func(s Session) error
+type RawSessionDialer func(legNumber uint32) (RawSession, error)
+type ReconnectCallback func(s Session, r RawSession, legNumber uint32) (int, error)
 
-// Establish a Session that reconnects across temporary network failures. The
+// Establish Session(s) that reconnect across temporary network failures. The
 // returned Session object uses the given dialer to reconnect whenever Accept
 // would have failed with a temporary error. When a reconnecting session is
 // re-established, it reissues the Auth call and Listen calls for each tunnel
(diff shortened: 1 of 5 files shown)


session.go
```diff
@@ -663,14 +706,39 @@ func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {
...
 		}
 
 		auth.Cookie = resp.Extra.Cookie
-		return nil
+
+		// store any connect server addresses for use in subsequent legs
+		if cfg.EnableMultiLeg && legNumber == 0 && len(resp.Extra.ConnectAddresses) > 1 {
+			testServerAddr := strings.Replace(cfg.ServerAddr, ".lan", "", 1)
+			testServerAddr = strings.Replace(testServerAddr, ".dev-", ".", 1)
+			testServerAddr = strings.Replace(testServerAddr, ".stage-", ".", 1)
```
natasha-jarus, line 721: test code?

internal/tunnel/client/reconnecting.go
```diff
@@ -284,33 +363,79 @@ func (s *reconnectingSession) connect(acceptErr error) error {
...
-		s.Info("client session established")
-		s.stateChanges <- nil
+			s.Info("client session established")
+			s.stateChanges <- nil
+		}
 		return nil
 	}
 }
+
+func reconnectTunnelToSession(raw RawSession, t *tunnel, newTunnels map[string]*tunnel, oldID string) error {
```
natasha-jarus, line 409: i feel like this should be a method on the reconnecting session?

session.go
bobzilladev: i meant "test" there as in testing equality with default hostname, but it does bake some of our testing environment naming in here, which isn't the best. it needs to know if this is a least-latency default address vs a customer-specified one, and we set the server_addr for those environments concretely. maybe it should be generalized to a regex.

internal/tunnel/client/reconnecting.go
bobzilladev: it didn't need that state so i didn't attach it, but yeah no one else should be calling this, will add 👍

session.go
```diff
@@ -909,6 +977,13 @@ func (s *sessionImpl) Heartbeat() (time.Duration, error) {
 func (s *sessionImpl) Latency() <-chan time.Duration {
 	return s.inner().Latency()
 }
+func (s *sessionImpl) ConnectAddreses() []ConnectAddress {
```
jrobsonchase, line 980:
```suggestion
func (s *sessionImpl) ConnectAddresses() []ConnectAddress {
```

internal/tunnel/client/reconnecting.go
```diff
@@ -284,33 +363,79 @@ func (s *reconnectingSession) connect(acceptErr error) error {
...
 		}
 
+		// check if more sessions need to be established
+		sendStateChange := true
+		if desiredLegs > len(s.sessions) {
+			// set up the next connection
+			s.createTunnelClientSession(s)
+			// not done with initial setup yet
+			sendStateChange = false
+		}
```
jrobsonchase, line 389: Can we go ahead and make this

```suggestion
		for i := len(sessions); i < desiredLegs; i++ {
			// set up the next connection
			s.createTunnelClientSession(s)
			// not done with initial setup yet
			sendStateChange = false
		}
```

for if/when we support >2 legs?

err, wait. If `createTunnelClientSession` calls `connect`, then I guess this block will get hit each time a new leg is opened, which will drive the creation of the rest of the legs. So no loop needed?

bobzilladev: aye, the intent was to chain the connects along, one finishing would start the next

session.go
```diff
@@ -43,6 +43,9 @@ func (avd *AgentVersionDeprecated) Error() string {
 	return (*proto.AgentVersionDeprecated)(avd).Error()
 }
 
+// ConnectAddress is a type wrapper for [proto.ConnectAddress]
+type ConnectAddress proto.ConnectAddress
```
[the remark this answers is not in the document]
bobzilladev: This is following the pattern of the `AgentVersionDeprecated`, which isn't used on the public api but is used on the agent side (`ifx/session.go`).

```
// AgentVersionDeprecated is a type wrapper for [proto.AgentVersionDeprecated]
type AgentVersionDeprecated proto.AgentVersionDeprecated
```

session.go
bobzilladev: I switched this to a regex to avoid the replace awkwardness.

internal/tunnel/client/reconnecting.go
```diff
@@ -122,67 +122,174 @@ type ReconnectCallback func(s Session) error
...
+
+func (s *reconnectingSession) SrvInfo() (resp proto.SrvInfoResp, err error) {
+	if sess := s.firstSession(); sess != nil {
+		return sess.SrvInfo()
+	}
+	return proto.SrvInfoResp{}, ErrSessionNotReady
+}
+
+func (s *reconnectingSession) ListenHTTP(opts *proto.HTTPEndpoint, extra proto.BindExtra, forwardsTo string, forwardsProto string) (Tunnel, error) {
+	return s.Listen("http", opts, extra, forwardsTo, forwardsProto)
```
OfTheDelmer, line 221: ahh yes, the forwardsProto 😂

session.go
```diff
@@ -287,6 +296,28 @@ func WithServer(addr string) ConnectOption {
...
+// WithAdditionalServers configures the network address to dial to connect to the ngrok
+// service on secondary legs. Use this option only if you are connecting to a custom agent
+// ingress, and have enabled multi leg.
+//
+// See the [server_addr parameter in the ngrok docs] for additional details.
+//
+// [server_addr parameter in the ngrok docs]: https://ngrok.com/docs/ngrok-agent/config#server_addr
+func WithAdditionalServers(addrs []string) ConnectOption {
+	return func(cfg *connectConfig) {
+		cfg.AdditionalServerAddrs = addrs
```
OfTheDelmer, line 308: Would it be too surprising to `cfg.EnableMultiLeg = true` here too?

bobzilladev: hrm, I think I'd lean towards keeping it an explicit on or off. I can imagine user code where the addressing is set by a different part of the codebase than something doing the multi-leg decision, and that indeed might be a surprising side-effect to have it auto-enable itself.

session.go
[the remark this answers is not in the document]
bobzilladev: Have successfully moved to `ConnectAddresses() []struct{ Region, ServerAddr string }` to remove that public type, thanks for the proof-of-concept 💯

jrobsonchase (approved):
:shipit:

PR #163, opened 2024-05-28
Propagate half-closes correctly in forward

euank:
Before, the following would not work as you would expect:

```go
// # One terminal
// $ ncat --recv-only -l 9090

// ngrok-go code
fwd, err := sess.ListenAndForward(
  ctx,
  "127.0.0.1:9090",
  config.TCPEndpoint(),
)

// fwd.URL() is 0.tcp.jp.ngrok.io:14517 for this example

// another terminal
// $ ncat --send-only 0.tcp.jp.ngrok.io 14517 < hello-world.txt
```

What we would expect from the above would be for the send side to send "hello world" and exit, and then the recv side to print "hello world" and also exit.

This is what happens if you do `ncat --send-only localhost 9090` instead of copying through the ngrok tcp tunnel.

Before this change, when copying through ngrok the recv side would not exit because the 'Close' of the connection did not get propagated through the 'join'.

I've also added a unit test showing this.

Thank you to @abakum for originally noticing this issue and offering a fix over in #137.
In the hopes of landing this more quickly, I've written a new version, derived from the internal ngrok agent's join code, which should thus be easier to review etc.

 To try and give credit correctly, I've maintained the original commit from #137 as well.

### DIFF
--- a/forward.go
+++ b/forward.go
@@ -53,18 +53,20 @@ func (fwd *forwarder) Wait() error {
 // compile-time check that we're implementing the proper interface
 var _ Forwarder = (*forwarder)(nil)
 
-func join(ctx context.Context, left, right io.ReadWriter) {
-	g := &sync.WaitGroup{}
-	g.Add(2)
-	go func() {
-		_, _ = io.Copy(left, right)
-		g.Done()
-	}()
-	go func() {
-		_, _ = io.Copy(right, left)
-		g.Done()
-	}()
+func join(ctx context.Context, left, right net.Conn) {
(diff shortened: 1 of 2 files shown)


bobzilladev (approved):
lgtm, and does line up with the internal join.

bobzilladev:
Ah, need the authtoken to run tests, will work out of the box as a branch on `ngrok` org if moved there.

euank:
Thanks for the review; I created a dummy PR over in #165 to verify it's happy wrt CI!
It would be nice if we could figure something out to allow external PRs to get proper CI too. Regardless, merging based on review + CI here

PR #164, opened 2024-05-28
replace log adapter module license symlinks with full files

masonj5n:
It seems like the license symlinks in the log adapter modules aren't being picked up by pkg.go.dev and preventing the docs from being displayed: https://pkg.go.dev/golang.ngrok.com/ngrok/log/slog

Seems to be a known issue:
https://github.com/golang/go/issues/40586


bobzilladev (approved):
Looks to match the main license, thanks for tackling

PR #165, opened 2024-05-29
(Duplicate for CI) Propagate half-closes correctly in forward

euank:
Duplicate of #163, but as a branch on this repo, to force the build-and-test job to run once.

This is the same code, so if it passes here, the intent is to close this PR and merge the original.
Due to using an authtoken secret, we need a local branch + pull request in order to actually run the CI step, and while I'm fairly confident this should pass CI, it's worth verifying properly before merging, even if juggling branches like this is a bit inconvenient.


euank:
Passed CI, cool, closing so I can go back and merge the original.

PR #166, opened 2024-06-11
policy: erase type to any

jrobsonchase:
We're probably going to be making more changes to the schema of this. Guard against having to release new SDKs all the time whenever we do so, and defer validation to the server side.

~Note: this is currently a *breaking* change - probably want to make it backwards-compatible.~
OK, backwards-compatible now! `any` is better than `map[string]any` for this.

### DIFF
--- a/config/common.go
+++ b/config/common.go
@@ -19,7 +19,7 @@ type commonOpts struct {
 	ForwardsProto string
 	// Policy that define rules that should be applied to incoming or outgoing
 	// connections to the edge.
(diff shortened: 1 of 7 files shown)


jrobsonchase:
Hmmmm, this passes tests, but the `Action.Config` type has changed on the wire, so this won't actually work. Will need server-side support to fully support this new approach.

CK-Ward:
Closing this per today's discussion. @nijikokun @tammybailey

PR #167, opened 2024-06-20
Add WithBindings Mixin

Megalonia:
Support WithBindings() mixin for all agent endpoint types

### DIFF
--- a/config/bindings.go
+++ b/config/bindings.go
@@ -1,0 +1,30 @@
+package config
+
+type bindings []string
+
+// WithBinding configures ingress for an endpoint
+//
+// The requestedBindings argument specifies the type of ingress for the endpoint.
+func WithBindings(requestedBindings ...string) interface {
+	HTTPEndpointOption
+	TLSEndpointOption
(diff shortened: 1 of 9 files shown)


config/bindings.go
```diff
@@ -0,0 +1,35 @@
...
+
+func (b bindings) ApplyTCP(cfg *tcpOptions) {
+	cfg.Bindings = []string(b)
+}
+
+func (b bindings) ApplyHTTP(cfg *httpOptions) {
+	cfg.Bindings = []string(b)
+}
+
+func (b bindings) ApplyLabeled(cfg *labeledOptions) {
```
Megalonia, line 33: do we need this for labeled tunnels? was just following what other BindExtra fields did

euank: My understanding is that we don't support bindings for labeled tunnels, so this should be removed, thanks for checking!

euank (approved):
Code-wise looks good to me, though I think the git log would be more readable with the commits squashed down before merging!

PR #170, opened 2024-06-25
add support for TrafficPolicy

TheConcierge:
### Why 
This is part of an effort to remove strict PolicyEngine definitions from our public SDKs. By doing so, we can rapidly iterate changes to the PolicyEngine without having to update client definitions every single time. The backend has already been updated to support policy-as-string, so now the SDK needs to be updated.

#### Note
The required backend changes are still under review and are not yet part of production ngrok.

### How
The protobuf definition was updated to contain a new `TrafficPolicy` field of type `string`. In the `config/policy.go` file, a new type was defined on top of `string`, adding functionality that adheres to the various required interfaces for our various connection types. These functions do nothing more than populate the new `TrafficPolicy` field with the provided string representation, which will eventually get passed to the appropriate backend service.

Additionally, to support backwards compatibility, `WithPolicy` was updated to convert the original, structured definition into an appropriate string. While this won't work forever, it should allow users who upgrade their SDK version to not _have_ to change anything until we do the larger, phased-based cutover in the future. A printout was added to warn users about the upcoming deprecation of `WithPolicy`.

### Validation
Unit tests were updated, checking that:
- YAML policy strings are validated
- JSON policy strings are validated
- policies specified by `WithPolicy` are properly converted to a policy string and placed in `TrafficPolicy`.

Tunnels were started using both `WithPolicy` and `WithPolicyString`. Both started without error and applied the correct policy.

https://github.com/ngrok/ngrok-go/assets/10949861/108991a2-4a16-49ac-ac73-b19bdd89b16f


config/common.go
```diff
@@ -17,9 +17,12 @@ type commonOpts struct {
 	// Currently only relevant for HTTP/1.1 vs HTTP/2, since there's a potential
 	// change-of-protocol happening at our edge.
 	ForwardsProto string
+
+	// DEPRECATED: use TrafficPolicy instead.
+	Policy *policy
```
nikolay-ngrok, line 22: Is this exposed in the config? Should we somehow pull it in `TrafficPolicy`?

TheConcierge: im not sure i understand the question.
is the concern that somebody might manually set Policy on the config struct vs using WithPolicy?
if that's a concern, we can always update the ApplyTCP, ApplyTLS, and ApplyHTTP functions to do the conversion instead. @nikolay-ngrok lmk if that would address the concern. I'm honestly leaning towards doing that anyway 🤷

nikolay-ngrok: yeah, if ppl are passing in this field, they would expect it to work even if deprecated? or we should just return an error or something to make it clear.

TheConcierge: good call, my currently solution does not address all cases. I'll revert the `WithPolicy` function and change the implementations of `Apply<Proto>` to:
- convert `Policy` to `TrafficPolicy`
- print out the Deprecation warning 

This covers the cases where:
- the user uses `WithPolicy`
- the user creates the `Policy` struct without helpers and adds it to the config

TheConcierge: ok, change was made. Tests still run, policy still works on tunnel using old way of configuration

nikolay-ngrok (approved):
👍

PR #182, opened 2024-08-28
feat(ci): Add check for go mod tidy

jonstacks:
Adds a check to make sure that `go mod tidy` produces no diff for the `go.mod` or `go.sum` files.

Also updates some of our actions versions while we are here.


jonstacks:
Looks like tests don't work from a fork :thinking: Going to push this up to the repo and try again

PR #183, opened 2024-08-28
feat(ci): Add check for go mod tidy

jonstacks:
Supersedes #182 

Adds a check to make sure that go mod tidy produces no diff for the go.mod or go.sum files.

Also updates some of our actions versions while we are here.


jonstacks:
Validation can be seen in the first commit for adding the new check. `go.mod` wasn't tidy and so the added check fails. We fix it in the second commit.

nikolay-ngrok (approved):
👍

PR #185, opened 2024-08-28
fix: run go mod tidy again

jonstacks:
Status checks didn't run on https://github.com/ngrok/ngrok-go/pull/184, but they should be running now


nikolay-ngrok (approved):
👍 👏

PR #189, opened 2024-09-20
config: allows_pooling boolean defaults false

warrn:


config/allows_pooling.go
```diff
@@ -0,0 +1,28 @@
+package config
+
+type allowsPoolingOption bool
+
+func WithAllowsPooling(allowsPooling bool) interface {
+	HTTPEndpointOption
+	TCPEndpointOption
+	TLSEndpointOption
+	LabeledTunnelOption
```
warrn, line 9: Have to do it for labeled tunnel too because common opts... we can have a different struct which is with is `allButLabeledCommon` if you prefer. Or I can just break it out for the 3.

We did it this way for description for agent endpoints for example

jrobsonchase: Just because it's in the `CommonOpts` doesn't mean we have to expose the `LabeledTunnelOption` interface. We can just let it be a dead option for labeled tunnels.

```suggestion
```

config/labeled.go
```diff
@@ -66,8 +66,9 @@ func (cfg *labeledOptions) WithForwardsTo(url *url.URL) {
 
 func (cfg labeledOptions) Extra() proto.BindExtra {
 	return proto.BindExtra{
-		Metadata:    cfg.Metadata,
-		Description: cfg.Description,
+		Metadata:      cfg.Metadata,
+		Description:   cfg.Description,
+		AllowsPooling: cfg.AllowsPooling,
```
jrobsonchase, line 71:
```suggestion
```

jrobsonchase (approved):
Let's drop the `LabeledTunnelOption` interface since it won't be supported there, but otherwise lgtm :+1:

config/allows_pooling.go
warrn: There was a situation where leaving an option undefined broke labeled tunnels with description... I don't recall what so I'll look.

warrn: Okay I figured out what happened:

https://github.com/ngrok-private/ngrok/blob/main/go/cmd/ngrok/tung/config.go#L57

We do this for every tunnel, including LabeledTunnels... as long as we don't call code this way for allows_pooling it'll be fine.

jrobsonchase: Yeah, let's just do it the right way from the agent side so we don't have to pretend to support pooling with labeled tunnels.

jrobsonchase (approved):

PR #191, opened 2024-10-09
update changelog for v1.11.0 release

TheConcierge:
This captures the rest of the changes since the `v1.10.0` release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
 ## 1.11.0
 
+Enhancements:
+
+- Adds `WithAllowsPooling` option, allowing the endpoint to pool with other endpoints with the same host/port/binding
+- Adds `WithURL` option, specifying a URL for the endpoint
+- Adds `WithTrafficPolicy` option, applying Traffic Policy to the endpoint
+
 Changes:
 
-- Adds `WithTrafficPolicy` function, replacing `WithPolicyString`
+- Deprecates `WithPolicyString` in favor of `WithTrafficPolicy`
 
 ## 1.10.0
 


CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
 ## 1.11.0
 
+Enhancements:
+
+- Adds `WithAllowsPooling` option
+- Adds `WithURL` option
```
jrobsonchase, line 6: Might be nice to have a little blurb about what each of these actually do. 

They don't actually have godocs at the moment, which is also a problem, but probably not one for this change.

CHANGELOG.md
```diff
@@ -1,8 +1,13 @@
...
+
 Changes:
 
-- Adds `WithTrafficPolicy` function, replacing `WithPolicyString`
+- Replaces `WithPolicyString` option with `WithTrafficPolicy`
```
jrobsonchase, line 10: Can we change the wording of this to indicate that `WithPolicyString` is just _deprecated_ and not replaced entirely? Was about to come in hot with the "no breaking changes without a major version" hammer, but then discovered that that isn't actually the case :sweat_smile:

CHANGELOG.md
TheConcierge: this would be nice! however, i have no idea what they do!

CHANGELOG.md
TheConcierge: yeah you're right, this wording isn't so good. I'll probably move `WithTrafficPolicy` to enhancements and leave a deprecation message for `WithPolicyString`

CHANGELOG.md
TheConcierge: I reached out to the endpoints project to see if anybody had anything additional, but got no response. I may need to nudge them again

jrobsonchase (approved):

PR #196, opened 2024-12-19
Move proto to dedicated dir ; Add agent/conn_header.go proto

hjkatz:
# Why

* The agent conn_header is something we want to use in the ngrok-operator. Rather than vendoring the result there we choose to expose the proto in our SDK then consume in the client (ngrok-operator).
* I thought to move the proto around and group them into `proto/`

# How

* Move proto to `proto/`
* Update `scripts/Makefile`
* Add `proto/conn_header.go` in `agent` package
* Rename `internal/pb` to `internal/mw`


hjkatz:
:(

PR #198, opened 2025-01-22
endpoints bind to a single binding 

Megalonia:
moving towards 1 binding per endpoint


config/bindings.go
```diff
@@ -1,30 +1,26 @@
 package config
 
-type bindings []string
+type bindings string
```
Megalonia, line 3: needs to be singular

rkolavo:
believe you'd want a changelog

Abdiramen (approved):

Issue #193, opened 2024-10-10
Document things that are missing godocs


jrobsonchase:
I'm currently aware of `WithAllowsPooling` and `WithURL`, but there are probably more. They should all be revealed by #192.

Issue #178, opened 2024-08-14
Session.ListenAndHandleHTTP has an awkward to use type signature


euank:
### Problem

The type signature for `ListenAndHandleHTTP` takes a pointer to an interface (`*http.Handler`) for the handler instead of just the interface directly.

This is awkward because, for example, the following errors out:

```go
ngrok.ListenAndHandleHTTP(
	context.Background(),
	config.HTTPEndpoint(),
	http.HandlerFunc(func(rw http.ResponseWriter, req *http.Request) {
	}),
)

// cannot use config.HTTPEndpoint() (value of type config.Tunnel) as *"net/http".Handler value in argument to ngrok.ListenAndHandleHTTP: config.Tunnel does not implement *"net/http".Handler (type *"net/http".Handler is pointer to interface, not interface)
```

I believe it was intended for that last argument to be `http.Handler`, just like the stdlib [http.ListenAndServe](https://pkg.go.dev/net/http#ListenAndServe) takes.

### What next

So, how do we solve this?

Well, in my opinion we update it to not be a pointer, but my main question is one of backwards compatibility. Technically that's a breaking change, so do we:

1. Release v2 (`golang.ngrok.com/ngrok/v2`)
2. Add a new method, deprecate the old (`ngrok.ListenAndHandleHTTPReal` or whatever, naming is hard)
3. Update it in place, hope no one's using it

It's unfortunate that the v1 to v2 jump in a go module makes the import path more confusing looking (with 'ngrok' no longer being the trailing part), but I figure we'll have to do it eventually anyway, so maybe that's the right answer?

I figured I'd ask to see if we have any opinions on this, happy to put up a PR for whichever if we have consensus on the path!

jrobsonchase:
Well that's annoying. Probably some bad copy-pasta.

As far as fixing it goes, I hate all of those options :sob: But we're actively working on a new API for v2, so we could just wait until we're ready to release it and address this issue at that time.

Another (grosser, imo) option may be to replace the argument type with `any`, and then do some type assertions to do the right thing in the method body?

inconshreveable:
this API no longer exists in v2, closing

PR #203, opened 2025-05-04
v2 API

### LINKED ISSUE #202: Using the option config.WithURL should infer the scheme
## What I expected to happen

When making a `config.Tunnel` that looks like the following:

```go
tunnelConfig := config.HTTPEndpoint(
        config.WithURL("http://foo.bar.internal:80"),
)
```

I expected the scheme to automatically be inferred as "http". 

## What actually happened

It ended up defaulting the endpoint scheme to "https" and the endpoint showed up as a https endpoint in the ngrok dashboard.

I then had to also pass:

```go
tunnelConfig := config.HTTPEndpoint(
        config.WithURL("http://foo.bar.internal:80"),
        config.WithScheme("http"), // This had to be added
)
```

In order to get it to work

## Proposed change

We should make it so that if you supply a scheme in the URL, it is respected and you don't have to also specify it.

inconshreveable:
v2 API for ngrok-go

- Dramatically simplified the API to remove many overlapping options, options that are now deprecated, unnecessary convenience functions, etc.
- Simplified the API by creating a unified API by removing all protocol-specific behaviors (which have all been moved to Traffic Policy).
- Removed the config package. All of its options are now folded into the top-level package or removed because they were migrated into Traffic Policy.
- Updates the API to use new ngrok terminology of Agents, Endpoints and Traffic Policy.
- Removes unnecessary functionality that has been moved to Traffic Policy.
- Removes functionality that is now deprecated (like labeled tunnels).
- Added support for agent-based TLS termination and Mutual TLS termination.
- Added support for full TLS control over forwarding to the upstream
- Removed the prototype policy package that was not well supported.
- Separated out a concept of an Agent from its Session which were previously co-mingled.

Summarized new API:
```
package ngrok // import "golang.ngrok.com/ngrok/v2"

var DefaultAgent = NewAgent(WithAuthtoken(os.Getenv("NGROK_AUTHTOKEN")))
func Forward(ctx context.Context, upstream *Upstream, opts ...EndpointOption) (EndpointForwarder, error)
func Listen(ctx context.Context, opts ...EndpointOption) (EndpointListener, error)
func NewAgent(agentOpts ...AgentOption) (Agent, error)
type Agent interface{ ... }
type AgentOption func(*agentOpts)
    func WithAgentConnectCAs(pool *x509.CertPool) AgentOption
    func WithAgentConnectURL(addr string) AgentOption
    func WithAgentDescription(desc string) AgentOption
    func WithAgentMetadata(meta string) AgentOption
    func WithAuthtoken(token string) AgentOption
    func WithAutoSessionManagement(auto bool) AgentOption
    func WithClientInfo(clientType, version string, comments ...string) AgentOption
    func WithDialer(dialer Dialer) AgentOption
    func WithEventHandler(handler EventHandler) AgentOption
    func WithHeartbeatInterval(interval time.Duration) AgentOption
    func WithHeartbeatTolerance(tolerance time.Duration) AgentOption
    func WithLogger(logger log.Logger) AgentOption
    func WithMultiLeg(enable bool) AgentOption
    func WithProxyURL(urlSpec string) AgentOption
    func WithRPCHandler(handler RPCHandler) AgentOption
    func WithTLSConfig(tlsCustomizer func(*tls.Config)) AgentOption
type AgentSession interface{ ... }
type Dialer interface{ ... }
type Endpoint interface{ ... }
type EndpointForwarder interface{ ... }
type EndpointListener interface{ ... }
type EndpointOption func(*endpointOpts)
    func WithAgentTLSTermination(config *tls.Config) EndpointOption
    func WithBindings(bindings ...string) EndpointOption
    func WithDescription(desc string) EndpointOption
    func WithMetadata(meta string) EndpointOption
    func WithPoolingEnabled(pool bool) EndpointOption
    func WithProxyProtocol(proxyProtocol string) EndpointOption
    func WithTrafficPolicy(policy string) EndpointOption
    func WithURL(urlSpec string) EndpointOption
type Error interface{ ... }
type Event interface{ ... }
type EventAgentConnectSucceeded struct{ ... }
type EventAgentDisconnected struct{ ... }
type EventAgentHeartbeatReceived struct{ ... }
type EventHandler func(Event)
type EventType int
    const EventTypeAgentConnectSucceeded EventType = iota ...
type RPCHandler func(context.Context, AgentSession, rpc.Request) ([]byte, error)
type Upstream struct{ ... }
    func WithUpstream(addr string, opts ...UpstreamOption) *Upstream
type UpstreamOption func(*Upstream)
    func WithUpstreamProtocol(proto string) UpstreamOption
    func WithUpstreamTLSClientConfig(config *tls.Config) UpstreamOption
```

fixes #202 #193 #178


inconshreveable:
This is ready for review. A few remaining items I'd like to get through myself before merge:

1. Some additional test cases
1. A thorough read through to audit for anything I've missed in the authoring process
1. A git tag for the 2.0.0 release

endpoint.go
```diff
@@ -0,0 +1,126 @@
...
+	ID() string
+
+	// Metadata returns the endpoint's opaque user-defined metadata set by WithMetadata.
+	Metadata() string
+
+	// Protocol is sugar for URL().Scheme
+	Protocol() string
+
+	// ProxyProtocol returns the proxy protocol setting for the endpoint.
+	ProxyProtocol() string
```
nijikokun, line 44: We call `ProxyProtocol` in the configuration file `upstream.proxy_protocol` wondering if we should change this in either the configuration file or in here.

nijikokun (approved):
Wow, this feels really nice so far. Tests and formatting passed locally on macOS and Windows for me and no major issues in the code from what I read / tested. I had one note on one of the options.

endpoint_options.go
```diff
@@ -0,0 +1,310 @@
...
+func configureEndpoint(scheme string, endpointOpts *endpointOpts) (config.Tunnel, error) {
+	switch scheme {
+	case "http", "https":
+		return configureHTTPEndpoint(endpointOpts)
+	case "tcp":
+		return configureTCPEndpoint(endpointOpts)
+	case "tls":
+		return configureTLSEndpoint(endpointOpts)
+	default:
+		return nil, fmt.Errorf("unsupported URL scheme: %s", scheme)
```
Alice-Lilith, line 113: nit: It'd be nice to have the allowed options as enums or have the error message return the allowed strings so users trying it and getting it wrong know what is accepted more easily.

examples/forward/main.go
```diff
@@ -0,0 +1,38 @@
...
+	"context"
+	"fmt"
+	"log"
+	"os"
+
+	"golang.ngrok.com/ngrok/v2"
+)
+
+// example usage:
+// ./http https://mydomain.ngrok.app http://localhost:8080
```
Alice-Lilith, line 13: Did you mean `ngrok http --url=https://mydomain.ngrok.app http://localhost:8080` to show the equivalent cli command, or am I just not understanding this comment's syntax?

Alice-Lilith (approved):
👍 to @bmpngrok's nits, otherwise looks clean to me.

examples/forward/main.go
inconshreveable: it's actually the example way to call the executable created by this example. i suppose it should say `./forward ...` will fix and improve the comment to clarify

endpoint_options.go
inconshreveable: good call, ty, will fix. once we do a little more surgery on this we'll be able to remove this entirely. this switch is really only here to deal with the legacy API

examples/rpc/main.go
```diff
@@ -0,0 +1,72 @@
...
+	listener, err := agent.Listen(context.Background())
+	if err != nil {
+		log.Fatalf("Error creating endpoint: %v", err)
+	}
+	log.Printf("Endpoint created: %s", listener.URL())
+
+	// Wait for signal to exit
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+	<-sigChan
```
euank, line 67: We're still stuck waiting here, even if we get a `Step` request in the dashboard.

We do get an INFO level log line (`level=INFO msg="accept failed" obj=csess id=958a03312932 err="reconnecting session closed"`) for the disconnect caused by clicking 'Stop' in the dashboard, but as best I can tell there's no way to observe it on the session object or agent object.

I wanted to do something like `<-agent.Wait(ctx)` or `<-session.Wait(ctx)` to fix it...

But since we don't have that, I think we'll have to make our own shutdown channel, and plumb closing that through the `WithRPCHandler`.

euank (commented):
The new package is way simpler, and I overall like the UX!

The only wart I've found so far is that waiting for a session to shutdown (i.e. from clicking 'stop' in the dashboard) doesn't seem possible with this API. I could be missing something, and I'd be happy to know if I am

I also have a small set of code nits

forwarder_test.go
```diff
@@ -0,0 +1,285 @@
...
+	return l.conn, nil
+}
+
+func (l *mockListener) Close() error {
+	l.closed = true
+	return nil
+}
+
+// TestForwarderBasic tests the basic forwarding functionality
+func TestForwarderBasic(t *testing.T) {
```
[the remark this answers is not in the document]
inconshreveable: ty for the catch! i've removed this test entirely and will rely on the forwarder integration tests to ensure there's coverage of this piece

forwarder_test.go
```diff
@@ -0,0 +1,285 @@
...
+	// Set a timeout for the entire test
+	testCtx, cancel := context.WithTimeout(context.Background(), 1*time.Second)
```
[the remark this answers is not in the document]
inconshreveable: removed this file, no longer relevant

forwarder_test.go
```diff
@@ -0,0 +1,285 @@
...
+		_, err := io.Copy(left, right)
+		t.Logf("Server->Client copy completed with err: %v", err)
+	}()
+
+	// Instead of blocking on WaitGroup, use a separate goroutine
+	// to avoid blocking the main test flow
+	go func() {
+		wg.Wait()
+		t.Log("Both copy operations completed")
+	}()
```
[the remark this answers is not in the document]
inconshreveable: removed this file, no longer relevant

inconshreveable:
> The only wart I've found so far is that waiting for a session to shutdown (i.e. from clicking 'stop' in the dashboard) doesn't seem possible with this API. I could be missing something, and I'd be happy to know if I am

Hmmm, that's true and I agree that it's a miss. On the other hand, it also was not possible with the previous version of the API. I'd like to advocate that we don't require that in this PR because this PR is already large enough and it can be safely added later without a breaking change. If that's amenable, I'll add it to the TODO

examples/rpc/main.go
inconshreveable: i've improved this example to show creating your own channel. see the comment below about an `agent.Wait()` or `session.Wait()` API

andrewryno:
I'm fairly certain this happens with v1 as well but this is my first time using the SDK so far so it was a fairly rough edge that would be great to address: if you don't have your authtoken env variable (or it's wrong), the SDK will just fail in the background and attempt to reconnect every 30 seconds and hang fover. I was only able to find out the issue by adding my own logger because it didn't even print it anywhere by default (I was running doing the connection in a `go test`).

examples/fasthttp/main.go
```diff
@@ -18,22 +17,19 @@ func main() {
...
-	tun, err := ngrok.Listen(ctx,
-		config.HTTPEndpoint(),
-		ngrok.WithAuthtokenFromEnv(),
-	)
+	ln, err := ngrok.Listen(ctx)
 	if err != nil {
 		return err
 	}
-	log.Println("tunnel created:", tun.URL())
+	log.Println("endpint online", ln.URL())
```
andrewryno, line 24:
```suggestion
	log.Println("endpoint online", ln.URL())
```

endpoint_options.go
```diff
@@ -0,0 +1,286 @@
...
+}
+
+// WithDescription sets a human-readable description for the endpoint.
+func WithDescription(desc string) EndpointOption {
+	return func(opts *endpointOpts) {
+		opts.description = desc
+	}
+}
+
+// WithMetadata sets opaque, machine-readable metadata for the endpoint.
```
andrewryno, line 59:
```suggestion
// WithMetadata sets an opaque, machine-readable metadata for the endpoint.
```

examples/multiple-endpoints/main.go
```diff
@@ -0,0 +1,140 @@
...
+	agent, err := ngrok.NewAgent()
+	if err != nil {
+		return err
+	}
+
+	// Connect the agent
+	if _, err := agent.Connect(ctx); err != nil {
+		return err
+	}
+	defer agent.Disconnect()
```
andrewryno, line 32: This is kind of one of the confusing parts to me about the new API where with the default agent you don't need to handle the connection/disconnection but can instead just directly call `Listen` or `Forward`. There's also no example where you use the resulting session, so having it return it and need to be ignored (in at least the ordinary case) seems a bit odd. This tripped me up a bit when configuring a custom agent.

Would it make sense to allow `Listen`/`Forward` connect if not already, and users can call `Connect` explicitly if they need to access the session information or tie the disconnection to some other signal than the default?

upstream.go
```diff
@@ -0,0 +1,65 @@
...
+	return func(o *Upstream) {
+		o.tlsClientConfig = config
+	}
+}
+
+// WithUpstreamProxyProto sets the PROXY protocol version to use when connecting
+// to the upstream server. Valid values are "v1" or "v2".
+//
+// See https://ngrok.com/docs/agent/config/v3/#upstreamproxy_protocol
+func WithUpstreamProxyProto(proxyProto string) UpstreamOption {
```
euank, line 53: Should we have an enum, like `ProxyProtocolV1` / `V2`, for this?

The string feels error-prone because both "1" and "v1" seem like reasonable inputs.

errors.go
```diff
@@ -1,164 +1,45 @@
...
-// Error arising from a failure to dial the ngrok server.
-type errSessionDial struct {
-	// The address to which a connection was attempted.
-	Addr string
-	// The underlying error.
-	Inner error
+func (e *errorAdapter) Code() string {
+	// Implementation based on the actual ngrok.Error API
+	// Since the actual implementation may differ, we're providing a simplified version
+	return "ERR_NGROK"
```
euank, line 25: It feels like this just needs to be `return e.ngrokErr.ErrorCode()`, right?

rpc/request.go
```diff
@@ -0,0 +1,18 @@
...
+)
+
+// Request defines the interface for handling RPC requests from the ngrok service.
+// It provides access to the method name and payload data.
+type Request interface {
+	// Method returns the RPC method name being called.
+	Method() string
+
+	// Payload returns the raw bytes of the request payload, if any.
+	Payload() []byte
```
euank, line 17: This UX feels off to me. We don't document anywhere what `Payload()` actually is, and so the user can't really do anything with it.

It's also never set anywhere in the code, so it feels premature to add it, this just always returns nil.

I feel like it would be cleaner to instead have a closed interface and type switch, so like to have the user do:

```go

type AgentRPC interface {
  closed()
}

type AgentRPCRestart struct {}
type AgentRPCStop struct {}

ngrok.WithRPCHandler(func(ctx context.Context, req AgentRPC) error {

   switch req.(type) {
     case AgentRPCStop:
     ....
   }

})
```

That leaves us with the ability to add a typed payload to future RPCs if we want, but also doesn't feel bad even if we never expand the set of rpcs there beyond update/stop/restart

euank (commented):
Another round of review!

I think the `RPC` bit needs another round of iteration, and I agree with ryno that the error code stuff should also be properly plumbed.

The `Agent` vs `AgentSession` also feels quite confusing, since a lot of the methods I'd expect to be on the `AgentSession` (like `Disconnect`) are actually on the Agent.

I get that the abstraction is "One agent has an automatically reconnecting session, so it can have multiple sessions over time meaning creating an endpoint should then re-create the endpoint, so it's tied to the agent not the session", but it still feels slightly off somehow.

errors.go
```diff
@@ -1,164 +1,45 @@
...
 }
```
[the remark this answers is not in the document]
inconshreveable: fixed and added a test to guarantee correctness

agent_options.go
```diff
@@ -0,0 +1,215 @@
...
+// See https://ngrok.com/docs/agent/#authtokens
+func WithAuthtoken(token string) AgentOption {
+	return func(opts *agentOpts) {
+		opts.authtoken = token
+		opts.sessionOpts = append(opts.sessionOpts, legacy.WithAuthtoken(token))
+	}
+}
+
+// WithAutoSessionManagement controls whether the agent automatically calls
+// Connect when needed or if the caller is responsible for this.
```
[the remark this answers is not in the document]
inconshreveable: indeed, updated the comment to be explain the behavior more thoroughly

examples/multiple-endpoints/main.go
inconshreveable: yeah, we're agreed. the intention was for `WithAutoSessionManagement()` to be default true which does exactly what you would expect, but that got missed. i'll update to make it default true and update the examples to no longer show explicit connection management.

i'm going to consider removing the return session from agent.Connect() as well. i think that makes sense too. you can just access it as .Session() after Connect completes and that seems good enough

rpc/request.go
inconshreveable: i don't want to create types for the RPC messages for two reasons:

1. This will allow callers to handle new RPC messages that get added without us needing to update every SDK
2. I believe there's a future where we allow our users to create and define their own RPC messages with their own RPC method names (and payloads)

`.Payload()` was a forward-thinking addition for point 2, but you're right that nothing is using it now and we can add it in a safely backwards-compat way when it does become relevant, so i'm going to drop that. ty!

forwarder.go
```diff
@@ -0,0 +1,203 @@
...
+	wg.Wait()
+}
+
+func (e *endpointForwarder) Close() error {
+	return e.CloseWithContext(context.Background())
+}
+
+func (e *endpointForwarder) CloseWithContext(ctx context.Context) error {
+	// Cancel the forwarding context to signal termination
+	e.cancel()
```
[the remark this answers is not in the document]
inconshreveable: agreed, these changes are implemented. cancellation has been moved into the forwardLoop so that when listener closes that all of the in-progress dials to backends and anywhere else we're using the context also stops

upstream.go
inconshreveable: resolved and fixed

PR #205, opened 2025-06-03
Remove all remaining log15 references, use slog.Logger everywhere

euank:
In the #203 PR we removed the public uses of `log15`, but still had a few references hanging around internally.

None of them seemed overtly difficult to remove, so I went through and cleaned them up.

This was mainly a rote replacement of `log15.Logger` to `*slog.Logger`, and `log15.Logger.New` to `slog.Logger.With` with only two exceptions:

1. We had some uses of `github.com/inconshreveable/log15/v3/ext.RandID()`. These became `fmt.Sprintf("%x", rand.Uint64())`, which I think gives us something pretty similar overall.
2. `netx.LoggedConn` embedded the `log15.Logger` interface; slog doesn't have an interface, so I manually added the correct `Info`, `Warn`, and `Error` method signatures instead

I also had to clean up the `examples` `go.mod` a little before `go mod tidy` there worked, but I think I got that right.

Overall, I think this is a straightforward refactor


inconshreveable (approved):
lgtm! ty for paying down some of the debt in the legacy package!

Issue #176, opened 2024-07-24
Program hangs where token is missing or is invaild without any feedback


swalx:
When starting an http endpoint, the program hangs (seems indefinitely) if the token is invalid or does not exists.

At first I thought simply checking if the environment variable exists and that it has a value would resolve the issue. However, this was not the case.
![image](https://github.com/user-attachments/assets/7014b0cc-eaba-45c7-af24-8dd54ce8a4c1)

I understand that the `Connect` func blocks until success but shouldn't it also return an error? In this case, the ngrok service should at least return an authentication error.

> // Connect begins a new ngrok [Session] by connecting to the ngrok service,
// retrying transient failures if they occur.
//
// Connect blocks until the session is successfully established or fails with
// an error that will not be retried. Customize session connection behavior
// with [ConnectOption] arguments.
func Connect(ctx context.Context, opts ...ConnectOption) (Session, error) {

Whether or not this is expected behaviour (for it to hang), I want to believe that where an attempt to start an http endpoint fails due to a missing or invalid token. The most appropriate one of the following errors should be returned to the user: [ERR_NGROK_105](https://ngrok.com/docs/errors/err_ngrok_105), [ERR_NGROK_106](https://ngrok.com/docs/errors/err_ngrok_106) or [ERR_NGROK_107](https://ngrok.com/docs/errors/err_ngrok_107). This would give the user enough feedback for them to make any necessary changes.

This can be reproduced by following the instructions exactly as seen in the [Quickstart example](https://github.com/ngrok/ngrok-go/blob/main/README.md#quickstart) found in the README with or without the example token.

This library is really awesome and useful. I just decided to test it out and realise this problem. If I am missing something, it would be great if I could be pointed in the right direction. Otherwise, I will continue to familiarise myself with the code base to potentially open a PR with a fix (assuming no one would have already done so).

Similar issue but includes the error: #107

CK-Ward:
Hi @swalx, thank you for bringing this to our attention. Are you still experiencing this issue? We'll have our team take a look.

jrobsonchase:
It's not quite that there's _no_ feedback, it's just somewhat hidden. Each time a connection is attempted and fails, or the session is disconnected, the `DisconnectHandler` is called with the error.

```go
	sess, err := ngrok.Connect(ctx,
		ngrok.WithAuthtokenFromEnv(),
		ngrok.WithDisconnectHandler(func(ctx context.Context, sess ngrok.Session, err error) {
[...]

PR #207, opened 2025-06-18
Update traffic policy example to work

euank:
The main issue was that we had a mix of spaces and tabs, and that makes yaml parsers unhappy.

Before:
```
2025/06/18 08:40:32 INFO client session established
2025/06/18 08:40:32 The TrafficPolicy you provided is invalid: could not decode from string: The traffic policy was not able to be parsed:
yaml: line 4: found a tab character that violates indentation.
```

After:
```
2025/06/18 08:43:52 INFO client session established
2025/06/18 08:43:52 Endpoint online https://<random>.ngrok.app
```

I also plumbed in slog since otherwise there's no output on errors, which is kinda unsettling.

### DIFF
--- a/examples/traffic-policy/main.go
+++ b/examples/traffic-policy/main.go
@@ -7,7 +7,9 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"log/slog"
 	"net/http"
+	"os"
 
 	"golang.ngrok.com/ngrok/v2"
 )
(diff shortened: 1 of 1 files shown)


inconshreveable:
hmmm, the test failures are for an unrelated CI reason that i'll need to track down. not sure what's going on there

when you say there was no output on errors, that's surprising, wouldn't that have occurred from the log.Fatal()? or is this because we're retrying behind the scenes and not returning an error?

euank:
The lack of output is because we're infinitely retrying errors, including invalid auth tokens and invalid traffic policy document errors. That's tracked over here: #176

euank:
> the test failures are for an unrelated CI reason

I accidentally made the PR from my fork of the repo rather than from a branch on the repo, so it can't use the repo's secrets. That's my bad.

I can confirm all those tests pass locally, so I think this is safe to merge

euank:
Re-created over in #208. Since we require CI for the merge button, it seems most pragmatic to split to a new PR, apologies for the extra noise!

PR #209, opened 2025-06-24
Add initial DisconnectHandler error handling to close session on unretryable error

ppeble:
# Summary

Addresses https://github.com/ngrok/ngrok-go/issues/188 and https://github.com/ngrok/ngrok-go/issues/176.

Adds in default disconnect handler behavior to immediately close sessions on 'non-retryable' errors, which for now is just auth errors. This will result in the following output using the quickstart code from the example:

```
10:28:36 phil@phil-System ngrok-test → go run main.go
2025/06/23 22:28:38 failed to connect: Usage of ngrok requires a verified account and authtoken.

Sign up for an account: https://dashboard.ngrok.com/signup
Install your authtoken: https://dashboard.ngrok.com/get-started/your-authtoken

ERR_NGROK_4018
exit status 1
```

Other errors will continue to retry indefinitely, as they do today.

# Details

I see that there was a large refactor pretty recently. Because of this I did not want to completely redo how the errors were done between the `legacy` and main package. 

This PR adds the `NonRetryable` interface in the `legacy` internal package and then tests for that behavior in the syntactic sugar function `Retryable(error)` in the `errors.go` in the root directory.

This way we can add additional `legacy` errors as `NonRetryable` by adding the appropriate function. When the decision to refactor away the `legacy` package comes hopefully the `Retryable` function can be reused and nothing calling it will have to change.


Alice-Lilith (approved):
@ppeble, thanks for the contribution and fixes 🙂

ppeble:
@Alice-Lilith I thought I had run all of these tests correctly locally with a valid auth token but maybe I didn't set the right env vars to run everything. Let me see if I can reproduce the failure locally later this afternoon

ppeble:
No, not at all! We can close this. As I mentioned I didn't want to do anything drastic because I saw the recent large refactor and figured new stuff would be coming anyways. I'll take a gander at the `EndpointOption` but don't wait on me if someone is ready to do it.

PR #210, opened 2025-07-14
Add event handler examples

jonstacks:
## What

This adds some examples for how to use an `EventHandler` that will get rendered into the pkgsite documentation. One is a basic example, the other includes a channel to 

## Why

It took me a while to figure out that the `Event`s I was getting in my custom `EventHandler`, were actually pointers to the `Event` types. I eventually found an example of how to use this in an integration test, but figured it would be helpful to surface this better.

image

### DIFF
--- a/events_test.go
+++ b/events_test.go
@@ -1,6 +1,8 @@
 package ngrok
 
 import (
+	"context"
+	"log/slog"
 	"testing"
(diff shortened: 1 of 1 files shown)


jonstacks:
Closing and re-opening for CI reasons.

PR #239, opened 2026-04-14
Remove unused `HandleTLSConnection` function

zombiezen:


bmpngrok (approved):
~~How'd this build? Or is the answer that it didn't but we didn't catch it?~~ I can't read

PR #240, opened 2026-04-14
Convert interfaces into concrete structs

zombiezen:
**This is a breaking change and should not be merged without a major version bump and a migration strategy from v2.**

Also fix race condition in `AgentSession`.

Part of AGENT-445

### DIFF
--- a/endpoint.go
+++ b/endpoint.go
@@ -1,65 +1,21 @@
 package ngrok
 
 import (
-	"context"
 	"crypto/tls"
 	"net/url"
 	"sync"
 )
 
-// Endpoint is the interface implemented by both EndpointListener and
-// EndpointForwarder.
+// Endpoint is the interface implemented by both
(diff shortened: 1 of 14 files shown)


endpoint.go
```diff
@@ -1,78 +1,22 @@
...
 	"net/url"
 	"sync"
 	"time"
 )
 
-// Endpoint is the interface implemented by both EndpointListener and
-// EndpointForwarder.
+// Endpoint is the interface implemented by both
+// [*EndpointListener] and [*EndpointForwarder].
 type Endpoint interface {
```
zombiezen, line 12: I was hacking on some other code over the weekend, and realized there's a better way of representing fixed sets of types like this: https://bsky.app/profile/zombiezen.com/post/3mjvmecvd5k26

(Reworking `Endpoint` to use this and implement some common methods is left as an exercise to the reader.)

PR #241, opened 2026-04-17
add support for patching agent endpoints

kmvachhani:
adds Update and UpdateUpstream to EndpointForwarder, allowing callers to mutate a running endpoint's metadata (name, description, metadata, pooling) and upstream URL without tearing down and rebinding the tunnel

### DIFF
--- a/agent.go
+++ b/agent.go
@@ -332,6 +332,18 @@ func (a *agent) ensureConnected(ctx context.Context) error {
 	return nil
 }
 
+// patchTunnelState sends a PatchTunnelState muxado message over the active
+// session to propagate metadata changes to the backend.
+func (a *agent) patchTunnelState(_ context.Context, tunnelID string, name, description, metadata *string, poolingEnabled *bool) error {
+	a.mu.RLock()
+	sess := a.sess
+	a.mu.RUnlock()
+	if sess == nil {
+		return fmt.Errorf("agent not connected")
+	}
+	return sess.PatchTunnelState(tunnelID, name, description, metadata, poolingEnabled)
+}
+
 // removeEndpoint removes an endpoint from the agent's list
 func (a *agent) removeEndpoint(endpoint Endpoint) {
 	// Remove the endpoint from our list under lock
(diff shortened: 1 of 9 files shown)


forwarder.go
```diff
@@ -23,6 +23,14 @@ type EndpointForwarder interface {
 	// UpstreamURL returns the URL that the endpoint forwards its traffic to.
 	UpstreamURL() url.URL
 
+	// UpdateUpstream atomically updates the upstream URL that connections are forwarded to.
```
bmpngrok, line 42: This breaks backcompat and will require us to cut a new Agent SDK version. Unless we want to just go ahead and fork the SDK now (I'm actually okay with that!), you'll need to do a one-two dance of introducing a new

```go
type UpdateableEndpointForwarder interface {
    ...
}
```

that ours happens to obey, and then cast as needed to get at it. You can look at `Diagnoser` to see this pattern in action

internal/tunnel/proto/msg.go
```diff
@@ -31,6 +31,9 @@ const (
 
 	// sent from client to the server
 	SrvInfoReq ReqType = 8
+
+	// sent from client to the server to patch mutable tunnel state
+	PatchTunnelStateReq ReqType = 10
```
bmpngrok: I think this can go up to the block around line 20?

agent.go
```diff
@@ -332,6 +332,16 @@ func (a *agent) ensureConnected(ctx context.Context) error {
 	return nil
 }
 
+func (a *agent) patchTunnelState(_ context.Context, tunnelID string, name, description, metadata *string, poolingEnabled *bool) error {
+	a.mu.RLock()
+	sess := a.sess
+	a.mu.RUnlock()
+	if sess == nil {
```
bmpngrok, line 339: Not your fault, but this is yet _another_ race condition with this design

bmpngrok (commented):
Some backcompat concerns with this, plus some minor code cleanup. If I'm missing something and the backcompat concern isn't valid, I can likely kick this to an Approve, but I figured we should talk first about it

forwarder.go
kmvachhani: ahhh yeah I'll add a new interface, thanks for catching this

agent.go
kmvachhani: yeah :/

internal/tunnel/client/reconnecting.go
```diff
@@ -213,6 +220,13 @@ func (s *reconnectingSession) listenTunnel(listen func(*session) (Tunnel, error)
 	return nil, ErrSessionNotReady
 }
 
+func (s *reconnectingSession) PatchTunnelState(tunnelID string, name, description, metadata *string, poolingEnabled *bool) error {
+	if sess := s.firstSession(); sess != nil {
```
bmpngrok, line 224: This should probably technically be locked to avoid the session disappearing. I think. I'm honestly okay with you punting it if we get onto a v3 Go SDK sooner rather than later

bmpngrok (approved):
I think you're good to ship this here

PR #246, opened 2026-06-04
privatedial: reconnect sessions after drain

eric-ngrok:
Adapt the private-dial client to keep a logical Session alive across server drain and control-stream failures in ngrok-go while preserving HTTP/2 and HTTP/3 transport selection.

- split Session supervision from per-transport connection state
- reconnect with backoff on drain or abrupt session errors
- preserve in-flight drained streams through server grace
- route new DialContext calls to the freshest usable transport
- use an owned h2 ClientConn instead of HTTP/2 transport pooling
- surface auth rejections as fatal instead of retrying forever

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -1,18 +1,22 @@
 // Package privatedial implements a client to allow dialing private ngrok
 // endpoints. It authenticates with an API Key as its authtoken and then multiplexes
 // per-target net.Conn streams over a single HTTP/2 or HTTP/3
-// connection.
+// connection. A Session transparently reconnects on server drain or abrupt
+// control-stream failure; new Dial calls follow the freshest underlying
+// transport while in-flight streams ride out the server-advertised drain grace
+// period on the old one.
 package privatedial
 
 import (
 	"bufio"
 	"context"
-	"crypto/rand"
+	cryptorand "crypto/rand"
 	"crypto/tls"
 	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
+	mathrand "math/rand/v2"
 	"net"
 	"net/http"
 	"net/url"
@@ -20,7 +24,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


euank (approved):
This is a mouthful of code for sure, but it looks good to me as an initial go at it!

I'm still running into one issue when using it in practice, which is that it takes way too long to actually reconnect. My manual testing looks like:

```
$ while sleep 1; do curl --connect-timeout=1s -s http://foo.internal; done

time=2026-06-08T22:54:00.364+09:00 level=DEBUG msg="dns: A" host=foo.internal ip=100.64.0.2
time=2026-06-08T22:54:00.364+09:00 level=DEBUG msg="dns: AAAA → empty NOERROR" host=foo.internal
time=2026-06-08T22:54:00.364+09:00 level=DEBUG msg="tcp forwarder request" dst=100.64.0.2:80
time=2026-06-08T22:54:00.378+09:00 level=DEBUG msg="tcp: proxying" host=foo.internal port=80
time=2026-06-08T22:54:00.403+09:00 level=DEBUG msg="copy half done" dir=local→remote bytes=76
time=2026-06-08T22:54:00.415+09:00 level=DEBUG msg="copy half done" dir=remote→local bytes=188
time=2026-06-08T22:54:01.422+09:00 level=DEBUG msg="dns: AAAA → empty NOERROR" host=foo.internal
time=2026-06-08T22:54:01.423+09:00 level=DEBUG msg="dns: A" host=foo.internal ip=100.64.0.2
time=2026-06-08T22:54:01.423+09:00 level=DEBUG msg="tcp forwarder request" dst=100.64.0.2:80
time=2026-06-08T22:54:01.434+09:00 level=DEBUG msg="tcp: proxying" host=foo.internal port=80

# successes every 1s, as expected

# at t=22:54:06, turn off network device
# at t=22:55:07, turn on network device again
time=2026-06-08T22:54:06.796+09:00 level=DEBUG msg="tcp forwarder request" dst=100.64.0.2:80
time=2026-06-08T22:54:50.801+09:00 level=WARN msg="tcp: dial failed" host=foo.internal port=80 err="dial tcp foo.internal:80: connect: connection refused"
```

So we can see it takes, in this case, 44 seconds to timeout and retry.
Note, this doesn't happen with http/3 (since http/3 survives a lot more connection abuse without even needing to create a new session, so the auto-reconnect isn't needed for this 1s network outage).
However, with http/2, it consistently takes quite a while before it reconnects, though it does eventually reconnect.

I think it's fair to merge this and iterate on it on the branch, since I do think it's overall right.

PR #247, opened 2026-06-08
privatedial: expose session's address

euank:
This exposes the remote address of the connection, which is useful for a client to log what server they connected to as a debug log.

It's helpful for us as we iterate on this to be able to debug things more quickly, and also lets us see if it's an ipv4 vs ipv6 connection, which is also sometimes helpful to know.

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -19,6 +19,7 @@ import (
 	mathrand "math/rand/v2"
 	"net"
 	"net/http"
+	"net/netip"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


eric-ngrok (approved):
Thanks!

eric-ngrok:
@euank I have rebased these changes here: https://github.com/ngrok/ngrok-go/commit/f74147a2b534996fafa5c66a818745aa02d2d716

If it looks good, go ahead and force reset this branch to that commit and merge!

PR #248, opened 2026-06-09
privatedial: update session stickyProto to be an atomic

euank:
It's easier to read this way, just a small cleanup tweak.


eric-ngrok (approved):
nice improvement!

PR #250, opened 2026-06-09
privatedial: switch error handling over to trailers

euank:
We originally speced out trailers as the mechanism for sending the error code / message, and then the initial MVP didn't do that but instead used http status codes and such.

The usual downside with http trailers is that they're not supported well in http/1, browsers, and some libraries, but this is an internal protocol where we control the client and server, so we can just insist on http/2 or 3 (as we do), and also only use libraries that deal with em correctly (as we do currently).

With those downsides out of the way, the upsides are quite nice since they let the server-side be simpler, and they're less ambiguous to parse client-side.

Let's switch over to em.

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -374,7 +374,20 @@ func (c *Client) race(ctx context.Context) (*Session, error) {
 
 	quicCancel()
 	h2Cancel()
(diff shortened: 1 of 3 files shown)


eric-ngrok (approved):
great!

PR #251, opened 2026-06-09
privatedial: speed up reconnect after network outage

eric-ngrok:
Fixes https://linear.app/ngrok/issue/GAT-338.

- bound each Session.dial by the dial-wait budget so a dial onto a silently-dead conn fails fast
- lower the HTTP/2 keepalive/PingTimeout

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -74,7 +74,15 @@ const (
 	// HTTP/2 attempt is staggered in. If QUIC completes within this window
 	// no second connection is ever made.
 	quicHeadStart = 250 * time.Millisecond
+
+	h2ReadIdleTimeout = 10 * time.Second
+	h2PingTimeout     = 5 * time.Second
+
+	quicKeepAlivePeriod = h2ReadIdleTimeout
+	quicMaxIdleTimeout  = h2ReadIdleTimeout + h2PingTimeout
 )
+
+var errDialWaitTimeout = errors.New("private-dial: dial wait timeout")
 
 // stickyProtocol records the first protocol the race settled on in this
 // process. Once set, every subsequent connect (including reconnects) reuses
@@ -553,8 +561,8 @@ func (d *Dialer) newH2Transport(ctx context.Context) (roundTripCloser, string, e
 	}
 	remoteAddr := tlsConn.RemoteAddr().String()
 	transport := &http2.Transport{
-		ReadIdleTimeout: 30 * time.Second,
-		PingTimeout:     15 * time.Second,
+		ReadIdleTimeout: h2ReadIdleTimeout,
+		PingTimeout:     h2PingTimeout,
 	}
 	cc, err := transport.NewClientConn(tlsConn)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


privatedial/client.go
```diff
@@ -794,11 +794,29 @@ func (s *Session) DialContext(ctx context.Context, network string, addr string)
 }
 
 func (s *Session) dial(ctx context.Context, addr dialAddr) (net.Conn, error) {
+	// budgetCtx bounds the entire dial — both the wait for a usable conn
+	// and the per-conn RoundTrip. A single goroutine arms the dial-wait
+	// timer and cancels budgetCtx when it fires, so a dial issued onto a
+	// silently-dead conn fails at the budget instead of blocking until the
+	// transport's own keepalive notices.
+	budgetCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
```
euank: I get that we're setting our own timer as a goroutine below so that we can inject a mock clock and do unit testing...

But I think this reads better as `context.WithTimeoutCause`, and using `context.WithTimeoutCause` doesn't require an extra goroutine (since internally that's a `time.AfterFunc`, which doesn't need a goroutine while sleeping), so it's more efficient too.

I think for unit testing this, we aren't getting _that much_ value out of it and I value the more efficient code slightly more...
Though ideally, we can shape the code such that `synctest` can unit test all the time stuff we care about.

privatedial/client.go
```diff
@@ -794,11 +794,29 @@ func (s *Session) DialContext(ctx context.Context, network string, addr string)
...
+
 	timer := s.sessionClock().NewTimer(s.dialWaitTimeout())
 	defer timer.Stop()
 
+	var timedOut atomic.Bool
```
euank: Rather than `timedOut`, we can use `WithTimeoutCause` / or `CancelCause` if we're using `WithCancel` still, plus a sentinel error value that lets us know we hit the timeout.

I think that reads cleaner.

privatedial/client.go
```diff
@@ -794,11 +794,29 @@ func (s *Session) DialContext(ctx context.Context, network string, addr string)
...
+		case <-timer.C():
+			timedOut.Store(true)
+			cancel()
+		case <-budgetCtx.Done():
+		}
+	}()
+
 	dialErrFor := func(err error) error {
-		if ctx.Err() == nil && errors.Is(err, context.DeadlineExceeded) {
+		if ctx.Err() == nil && (timedOut.Load() || errors.Is(err, context.DeadlineExceeded)) {
```
euank: Ditto on `CancelCause` being a cleaner way to differentiate errors here

euank (commented):
This overall looks good, though I've got some comments about how we juggle timeouts here!

Do we need similar changes for http/3?

privatedial/client.go
eric-ngrok: agreed on all of the above. Thanks for pointing me to `context.WithTimeoutCause` and `synctest`! Nifty

euank (approved):
This looks good, thank you for the improvements!

:shipit:

PR #253, opened 2026-06-10
privatedial: slim down interface a bit, simplify a bit, expose less, general polish

euank:
This does a mix of things that all fall under "polish" and "renaming".

First, let me list the very simple things which I think are pretty clearly straightforward:

1. Moved the protobuf types into `internal`, they're an internal impl detail
2. Removed unneeded methods (DrainCh, PingInterval, LastRTT) - I can see arguments for all of these, but I'd rather add them later than include them in the v1 interface
3. Remove ServerAddr / ServerName - Those predate adding QUIC, so we now have H2/QUIC prefixed fields instead
4. Updated comments to be more accurate

Now let's talk about the changes that involve a bit more "taste", and thus make more sense to talk through.

The big change is renaming from `Client` to `Dialer`, and removing `Session` entirely (with `Dialer` being the new `Session` object). The second big change is making the Dialer "lazy", and adding an explicit method to make it eager (`Connect()`)

For the rename, I think `privatedial.Client.DialContext` and `NewClient(ClientOpts)` read worse than `privatedial.Dialer.DialContext` and `privatedial.New(privatedial.Config{})`.

For the other change, folding `Session` into `Dialer` (ne `Client`), I need a little more justification.

First, `Client` before, in my opinion, was useless. It was a way to bundle config together and save it, but re-using the same `ClientOpts` (now `privatedial.Config`) by itself already accomplishes that, so it was an extra layer that wasn't needed.

Second, I think that `OpenSession` was the wrong affordance since it would eagerly dial, and I think it's convenient to be able to both lazy and eagerly dial.

To explain what I mean, the old UX looked like:

```go
clientOpts := privatedial.ClientOpts{...}
sess, err := NewClient(clientOpts).OpenSession(ctx)
// sess has connected, so 'err' here contains any session connect errors
sess.DialContext(ctx, ....)
```

The new UX has two supported paths, eager and lazy, which look like this:

```go
// Eager, closer to old path
pdConfig := privatedial.Config{...}
dialer := privatedial.New(pdConfig)
err := dialer.Connect(ctx)
// dialer has connected, so 'err' here contains any session connect errors, like "auth invalid"
conn, err := dialer.DialContext(ctx, "tcp", "foo.internal:80")
// err here will contain errors specific to this connection, like 'foo.internal does not exist'

// Lazy, impossible in old path
pdConfig := privatedial.Config{...}
dialer := privatedial.New(pdConfig)
conn, err := dialer.DialContext(ctx, "tcp", "foo.internal:80") // internally calls 'Connect'
// err here can be any session errors, like "auth invalid", but can also be dial errors, like "foo.internal"
```

The reason for this UX split is that we can make the lazy dial path faster (by putting auth + dial in one round trip), at the cost of the caller having to deal with more types of errors.
That cost seems small, so it seems reasonable to have to me.

It also means for things that don't need robust error handling (like once off scripts or whatever), you get to write less total code for the happy path, and that makes the package simpler to use in a common-enough case.

... Hopefully that's sorta compelling!

The other more notable change here is moving `ServerErrCh` to `Done() + Err()` methods, akin to how context works.
I think that's just cleaner, I can defend it more if anyone feels otherwise, but I feel like it doesn't need too many words around it.

Happy to break out any pieces of this that you'd like me to split out btw!


eric-ngrok (approved):
Changes make sense to me. Thank you!

PR #255, opened 2026-06-16
privatedial: fall back to HTTP/2 when QUIC can't get adequate UDP buffers

euank:
See https://github.com/quic-go/quic-go/wiki/UDP-Buffer-Sizes

Basically, there's known system configurations where quic is likely to be slower than tcp for large transfers, and in those cases it's probably best for us to default to tcp.

A user can force quic or h2 specifically if they know it's better for their environment, but for the default option I think this makes sense!


eric-ngrok (approved):
very nice, thank you

PR #256, opened 2026-06-16
privatedial: eagerly surface endpoint errors, and expose EndpointID

euank:
Before this change, it was difficult to correctly identify some errors, and there was no way to get the ID of the endpoint actually dialed.

This fixes them by, per the spec, adding a `DialResp` type which always gets read immediately.

If that type isn't written on the wire, that means it's an error, and the existing trailer mechanism is used to rehydrate the error.

The use of trailers for the error, rather than putting it in DialResp, is so that we can push down an error later (i.e. "agent went away") which happened after we finished the dial, and that mechanism being common to all errors seems a little cleaner to me.

This, of course, also requires server-side changes, see the internal repo for those.

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -856,14 +856,18 @@ func (d *Dialer) Err() error {
 //
 //	conn, err := dialer.DialContext(ctx, "tcp", "foo.internal:80")
 //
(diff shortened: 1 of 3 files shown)


eric-ngrok (approved):
awesome, this will be very helpful for the agent!

PR #257, opened 2026-06-22
privatedial: add /get-host endpoint for DNS resolution

eric-ngrok:
Fixes https://linear.app/ngrok/issue/AGENT-617

- Add Dialer.GetHost(ctx, host) (bool, error) implementing the unary /get-host endpoint, so a client can answer dig foo.internal with an A record or NXDOMAIN without dialing (no port required).
- Vendor the get-host protobuf messages (Error, GetHostReq/Resp, plus GetHostPortReq/Resp) by regenerating the private_dial stub.
- Interpret the in-band GetHostResp.Error: nil -> exists; the endpoint-not-found code (ERR_NGROK_706) -> NXDOMAIN; any other code -> a privatedial.Error, so auth/draining/transient failures are never mistaken for a negative DNS answer.
- Add a unaryRoundTrip helper for the unary /get-* path, reusing the existing auth, length-prefixed protobuf, stale-conn and trailer-error machinery.

### DIFF
--- a/privatedial/client.go
+++ b/privatedial/client.go
@@ -9,6 +9,7 @@ package privatedial
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	cryptorand "crypto/rand"
 	"crypto/tls"
@@ -34,6 +35,7 @@ import (
 	"github.com/quic-go/quic-go/http3"
 	"golang.org/x/net/http2"
 	"google.golang.org/protobuf/encoding/protodelim"
+	"google.golang.org/protobuf/proto"
 
 	pbpd "golang.ngrok.com/ngrok/privatedial/internal/pb_private_dial"
 )
@@ -520,6 +522,7 @@ func (d *Dialer) openConn(ctx context.Context, p Protocol) (*sessionConn, error)
 
 	controlURL := &url.URL{Scheme: "https", Host: serverAddr, Path: "/session"}
 	dialURL := &url.URL{Scheme: "https", Host: serverAddr, Path: "/dial"}
+	getHostURL := &url.URL{Scheme: "https", Host: serverAddr, Path: "/get-host"}
 
 	// io.Pipe for the request body — first frame is SessionReq, then
 	// the body is reused for client-→server ControlFrames (Ping/Pong).
@@ -606,6 +609,7 @@ func (d *Dialer) openConn(ctx context.Context, p Protocol) (*sessionConn, error)
 	h.log = h.log.With("server_id", h.serverID)
 	h.pingInterval = ack.GetPingInterval().AsDuration()
 	h.dialURL = dialURL
+	h.getHostURL = getHostURL
 	h.authToken = d.cfg.AuthToken
 	h.controlResp = resp
 	h.controlRespBody = respBody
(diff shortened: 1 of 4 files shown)


privatedial/client.go
```diff
@@ -983,6 +987,71 @@ func (d *Dialer) dial(ctx context.Context, addr dialAddr) (net.Conn, error) {
 	}
 }
 
+// GetHost reports whether host has a private endpoint reachable by this Dialer,
+// without dialing it. A nil error reports existence in the bool; a non-nil
+// error means the lookup failed (a privatedial.Error carries the ngrok code).
+// The result is racy: a later DialContext may still fail.
```
euank: I think the last sentence (that a Dial could still fail) isn't necessary, this would have to return an explicit handle of some sort for that sorta comment to be non-obvious imo.

I think we can also be a little cleaner in our description of the returns, i.e. something like:

```diff
- A nil error reports existence in the bool; a non-nil error means the lookup failed (a privatedial.Error carries the ngrok code)
+ It returns whether an endpoint exists, or an error. If a returned error has an ngrok error code associated with it, it will be of type [Error].
```

privatedial/client.go
```diff
@@ -1389,6 +1459,76 @@ func (h *sessionConn) dial(ctx context.Context, addr dialAddr) (net.Conn, error)
...
+	}
+
+	if !h.acceptsStreams() {
+		return &staleConnError{}
+	}
+	if reserver, ok := h.transport.(requestReserver); ok && !reserver.ReserveNewRequest() {
+		return &staleConnError{}
+	}
+
+	reqCtx, reqCancel := context.WithCancel(context.Background())
```
euank: I'm confused why this isn't just the ctx that gets passed in, like we're bounded by that context anyway, right?

privatedial/client.go
```diff
@@ -1389,6 +1459,76 @@ func (h *sessionConn) dial(ctx context.Context, addr dialAddr) (net.Conn, error)
...
+
+	type roundTripResult struct {
+		resp *http.Response
+		err  error
+	}
+	roundTripCh := make(chan roundTripResult, 1)
+	go func() {
+		r, err := h.transport.RoundTrip(reqWithContext)
+		roundTripCh <- roundTripResult{resp: r, err: err}
+	}()
```
euank: This all seems unnecessary, the channel is just to switch on the ctx or the req, but the request is already bounded by a context (which presumably should be the caller ctx), so as far as I can see we don't need a goroutine nor channel

privatedial/client.go
```diff
@@ -983,6 +987,71 @@ func (d *Dialer) dial(ctx context.Context, addr dialAddr) (net.Conn, error) {
...
+			return false, err
+		}
+		d.log.Debug("get-host raced a stale conn, retrying on a fresh one", "host", host, "error", err)
+	}
+}
+
+// interpretGetHost maps a GetHostResp's in-band Error to GetHost's result: nil
+// means it exists, the not-found code means it doesn't, and any other code is a
+// real failure surfaced as a privatedial.Error rather than a negative answer.
+func interpretGetHost(e *pbpd.Error) (bool, error) {
```
euank: It might just be me, but I feel like this being inlined at the one call-site reads cleaner than having a separate function for it.

"interpret" is also a kinda weird verb to me, and inlining also lets us avoid having to pick a function name here

euank (commented):
Overall looks pretty good!

There's one goroutine I don't understand, and I have a couple other small nits too

privatedial/client.go
eric-ngrok: thank you, agreed

privatedial/client.go
eric-ngrok: d'oh. fixed

privatedial/client.go
eric-ngrok: many thanks, fixed

privatedial/client.go
eric-ngrok: agreed, fixed

internal/integration_tests/listen_http_test.go
```diff
@@ -22,71 +19,8 @@ func TestListenAndHTTPRequest(t *testing.T) {
 	// Expected message
 	expectedMessage := "Hello, ngrok!"
 
-	// Create synchronization points
```
euank: These changes should be separate from this PR, right?

Like either they should be against main (if they're not there), or we should separately merge main into `privatedial` if we want to pick up new test changes for this branch.

euank (approved):
The privatedial changes LGTM! Thanks!

I think the `internal/integration_tests` changes should be off in a separate PR; approving contingent on that commit being split off.

PR #258, opened 2026-06-23
privatedial: add Config.IPProtocol to pin the server dial to IPv4/IPv6

euank:
Add an IPProtocol knob to Config that forces the underlying transport dial onto a single IP family. It controls both transports: the TCP socket for HTTP/2 and the UDP socket for HTTP/3 (QUIC). IPProtocolAny (the zero value) preserves today's dual-stack behavior.

This is a useful knob in networks where one family misbehaves or is slower.


eric-ngrok (approved):
Thanks!

PR #259, opened 2026-06-24
Reduce integration test HTTP flakes

eric-ngrok:
- Retry transient 404s while newly-created endpoints propagate.
- Add a shared listener request helper for HTTP and HTTPS tests.
- Use endpoint-ready requests in proxy protocol and URL pooling tests.

### DIFF
--- a/internal/integration_tests/test_utils.go
+++ b/internal/integration_tests/test_utils.go
@@ -4,6 +4,7 @@ import (
 	"bufio"
 	"bytes"
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"net"
@@ -17,6 +18,16 @@ import (
 	"golang.ngrok.com/ngrok/v2"
 	"golang.ngrok.com/ngrok/v2/internal/testcontext"
 )
+
+type httpStatusError struct {
+	URL        string
+	Status     string
+	StatusCode int
+}
+
+func (e *httpStatusError) Error() string {
+	return fmt.Sprintf("post %s: http %s", e.URL, e.Status)
+}
 
 // SkipIfOffline skips the test if NGROK_TEST_ONLINE environment variable is not set
 func SkipIfOffline(t *testing.T) {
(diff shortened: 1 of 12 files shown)


internal/integration_tests/test_utils.go
```diff
@@ -115,21 +126,112 @@ func MakeHTTPRequest(ctx context.Context, tb testing.TB, url string, message str
...
 	return resp, nil
 }
 
-// WaitForForwarderReady polls the forwarder endpoint until it responds or times out
+// MakeHTTPRequestWhenEndpointReady retries transient 404 responses that can
+// occur while a freshly-created endpoint is propagating through ngrok's edge.
+func MakeHTTPRequestWhenEndpointReady(ctx context.Context, tb testing.TB, url string, message string) (*http.Response, error) {
+	tb.Helper()
+
+	deadline := time.Now().Add(30 * time.Second)
```
kate-osborn: can we push this into the context (e.g. `context.WithTimeout()`)  so we only have to check `ctx.Err()` instead of both `ctx.Err()` and the deadline?

internal/integration_tests/test_utils.go
```diff
@@ -115,21 +126,112 @@ func MakeHTTPRequest(ctx context.Context, tb testing.TB, url string, message str
...
+		defer conn.Close()
+
+		message, err := handleHTTPRequest(conn)
+		if err != nil {
+			result <- httpListenerResult{err: fmt.Errorf("handle HTTP request: %w", err)}
+			return
+		}
+
+		result <- httpListenerResult{message: message}
+	}()
```
kate-osborn, line 195: probably not a huge deal for tests, but we can leak a goroutine here if we hit the timeout on line 206. we can address this by changing this function to accept a context and then spinning up another goroutine that closes the listener on ctx.Done().

kate-osborn (commented):
nice fixes! just a couple small comments

internal/integration_tests/test_utils.go
eric-ngrok: agree, fixed!

internal/integration_tests/test_utils.go
eric-ngrok: v nice, fixed

kate-osborn (approved):

